PR #9, opened 2019-12-05
add CD to kubernetes cluster

lwsanty:
1) TravisCI side:
- Docker image is built, pushed to GCP container registry
- image tag is sent to PubSub

2) GKE side, version controller:
- receives docker image tag
- updates web-app deployment's image tag


build-and-deploy.sh
```diff
@@ -0,0 +1,7 @@
+IMAGE=$1
```
lwsanty: add `set -e`

.travis.yml
```diff
@@ -5,9 +5,21 @@ node_js:
 os:
   - linux
 
+services:
+  - docker
```
dpordomingo: Do you have any idea about why this other build worked without declaring `docker` service?
https://travis-ci.org/dpordomingo/athenian-webapp/jobs/621079589

se7entyse7en: Seems that docker comes pre-installed in the Xenial build environment of travis. https://docs.travis-ci.com/user/reference/xenial/#docker

.travis.yml
```diff
@@ -20,3 +32,21 @@ jobs:
...
+        - if [ ! -d "$HOME/google-cloud-sdk/bin" ]; then rm -rf $HOME/google-cloud-sdk; export CLOUDSDK_CORE_DISABLE_PROMPTS=1; curl https://sdk.cloud.google.com | bash; fi
+        # Add gcloud to $PATH
+        - source /home/travis/google-cloud-sdk/path.bash.inc
+        - gcloud version
+        # Login to GCP
+        - echo ${GOOGLE_CREDENTIALS} | base64 --decode > google-credentials.json
+        - gcloud -q auth activate-service-account --key-file google-credentials.json
+        - gcloud config set project ${GOOGLE_PROJECT}
+        # Push
+        # TODO: need to discuss the tag
```
se7entyse7en: I think that the one you used it's totally fine until we have the first tagged version.

deploy/k8s/web-app/deployment.yml
```diff
@@ -0,0 +1,23 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: web-app-deployment
+  namespace: web-app
+spec:
+  replicas: 1
```
se7entyse7en: This means that there's only a pod running on the (currently 2 nodes) GKE cluster right?

deploy/k8s/web-app/service.yml
```diff
@@ -0,0 +1,19 @@
+kind: Namespace
+apiVersion: v1
+metadata:
+  name: web-app
+---
+# TODO consider changing service type to LoadBalancer to expose an endpoint
```
se7entyse7en: So I guess that this works now as expected because this matches a single pod, right? Do we have to change the `kind` to enable load balancing if we increase the number of pods?

deploy/k8s/web-app/deployment.yml
```diff
@@ -0,0 +1,23 @@
...
+      app: web-app
+  template:
+    metadata:
+      namespace: web-app
+      labels:
+        app: web-app
+    spec:
+      containers:
+        - name: web-app-server
+          image: gcr.io/athenian-1/web-app:v1
```
se7entyse7en: Is the image name fixed?

deploy/k8s/version-controller/deployment.yml
```diff
@@ -0,0 +1,50 @@
...
+      app: controller-app
+  template:
+    metadata:
+      namespace: web-app
+      labels:
+        app: controller-app
+    spec:
+      containers:
+        - name: controller-server
+          image: gcr.io/athenian-1/web-app-controller:latest
```
se7entyse7en: When is this pushed?

deploy/k8s/version-controller/cli/main.go
```diff
@@ -0,0 +1,130 @@
...
+	mc, err := newMetaClient(ctx, getConfig())
+	if err != nil {
+		log.Errorf(err, "failed to prepare metaClient")
+		os.Exit(1)
+	}
+
+	log.Infof("start receiving messages:")
+	for {
+		// TODO(@lwsanty): slack notification
+		if err := mc.subscription.Receive(ctx, func(ctx context.Context, msg *pubsub.Message) {
```
se7entyse7en: Is this blocking?

deploy/k8s/version-controller/cli/main.go
```diff
@@ -0,0 +1,130 @@
...
+			defer msg.Ack()
+			image := string(msg.Data)
+
+			log.Infof("got message: %q", image)
+			if err := mc.UpdateDeployment(image); err != nil {
+				log.Errorf(
```
se7entyse7en: Are we notified somehow? Not a problem right now, but just planning things we could improve.

se7entyse7en:
@lwsanty mostly questions related to k8s since I only know the concepts and don't have real experience.

deploy/k8s/version-controller/deployment.yml
```diff
@@ -0,0 +1,50 @@
+apiVersion: apps/v1
```
se7entyse7en: I guess that this has been deployed manually from the console?

se7entyse7en: What if it dies/exits?

build-and-deploy.sh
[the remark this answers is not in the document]
lwsanty: Done

.travis.yml
lwsanty: removed it since ^

.travis.yml
lwsanty: removed TODO

deploy/k8s/web-app/deployment.yml
lwsanty: yep, wanna change it? we could possibly add horizontal autoscale if needed...

deploy/k8s/web-app/service.yml
lwsanty: TBH there's a hack: when you change service type to LB - GKE handle it and assigns some dynamic IP address to access service externally

Since we have static IP address I could change this part to
```yaml
type: LoadBalancer
loadBalancerIP: *********
```

deploy/k8s/web-app/deployment.yml
lwsanty: Image name can be as it is, we have this tag in the registry, and it will be changed by CD anyways

deploy/k8s/version-controller/deployment.yml
lwsanty: now it's updated manually

deploy/k8s/version-controller/cli/main.go
lwsanty: I planned to have slack notifications in future

deploy/k8s/version-controller/deployment.yml
lwsanty: Yes, now this controller is being deployed manually

If pod dies it will be recreated by replication controller
If it exists I guess it will not be created due to name conflicts
If by any chance you run the same deployment by different name they will not conflict because only one pod will receive the message

deploy/k8s/version-controller/cli/main.go
lwsanty: you mean Receive method or Slack notifications?
I can't remember if receive was stucking
slack notifications - I can add them either now or later...

deploy/k8s/web-app/deployment.yml
se7entyse7en: no need know, just wanna understand 👍

deploy/k8s/web-app/service.yml
se7entyse7en: So GKE in case `type: LoadBalancer` would assign a dynamic a IP, and you change it manually using the static one that you created (I guess from the console) by overriding the `loadBalanceIP`. Did I understand right? And that IP is the one being pointed by the DNS using an A record, right?

deploy/k8s/version-controller/deployment.yml
se7entyse7en: Not for this PR, just thinking out loud. The idea of @dpordomingo of separating this whole thing could be a standalone project that for each new version just pushes its own image, so no need to push it manually. And then in this project, for example, we can simply have a config file with the version of the controller to use.

deploy/k8s/version-controller/cli/main.go
se7entyse7en: Nope not slack, I mean the `Receive` method.

deploy/k8s/web-app/deployment.yml
se7entyse7en: Sorry I didn't understand, shouldn't this match the image version pushed by Travis in the last build? I mean this: `${GOOGLE_DOCKER_IMAGE}:${TRAVIS_BRANCH}-${TRAVIS_COMMIT:0:7}`

deploy/k8s/version-controller/cli/main.go
lwsanty: Added delay as agreed

deploy/k8s/version-controller/deployment.yml
lwsanty: +1, I guess we even discussed it during the call

deploy/k8s/web-app/service.yml
lwsanty: yep, we could probably automate it somehow in future

deploy/k8s/web-app/deployment.yml
lwsanty: this deployment is used only during the very first deploy

an update operation does not need this entire config, it just updates the image field
https://github.com/athenianco/athenian-webapp/blob/6fe6808b67d37077e6d00216c81ae47be2985e7f/deploy/k8s/version-controller/cli/main.go#L119

deploy/k8s/web-app/service.yml
se7entyse7en: I've created an issue, just to remember.

PR #10, opened 2019-12-09
fix deploy script

lwsanty:
required by [ENG-38]

move all deploy section under deploy/deploy.sh file to prevent proceed on failures

[ENG-38]: https://athenianco.atlassian.net/browse/ENG-38


kuba--:
I thought about this kind of stuff (for different, personal project) and I decided that creating another hierarchy or specific file to deploy is not needed or won't be used.
So, it's up to you, but I thought about:
Either create a `Makefile` and move this functionality there so we'll have `make deploy`
or just create `travis.sh` and put all scripts related to CI there.

se7entyse7en:
> create a Makefile and move this functionality there so we'll have make deploy

I like this proposal, but not a blocker so up to you.

Makefile
```diff
@@ -0,0 +1,26 @@
+define deploy =
+set -e
+
+echo "preparing cloud"
+gcloud version || true
+if [ ! -d "$HOME/google-cloud-sdk/bin" ]; then rm -rf $HOME/google-cloud-sdk; export CLOUDSDK_CORE_DISABLE_PROMPTS=1; curl https://sdk.cloud.google.com | bash; fi
```
kuba--: For `if [ ! -d "$HOME/google-cloud-sdk/bin" ];` it should be `"${HOME}"` (otherwise it will just fetch the first `H`).

Makefile
```diff
@@ -0,0 +1,26 @@
...
+echo "adding gcloud to PATH env"
+source /home/travis/google-cloud-sdk/path.bash.inc
```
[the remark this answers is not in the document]
lwsanty: Done

Makefile
lwsanty: Done

lwsanty:
> Did you try to make it as a function - I think it's more readable, e.g.:
> 
> ```
> define deploy
> 	set -e
> 	echo "preparing cloud $(1)"
> 	if [ ! -d "${HOME}/google-cloud-sdk/bin" ]; then echo "NO - ${HOME}/google-cloud-
> sdk/bin"; fi
> # .... whatever ....
> endef
> 
> deploy:
> 	$(call deploy, Arg1, Arg2, Arg3)
> ```

TBH I don't see much readability win here, so IMO I would leave it as it is

kuba--:
As I said it's up to you. For me the main difference is:
`$(call deploy, Arg1, Arg2, Arg3)` vs  ` ; $(value deploy)`

kuba--:
I think it's a good idea to split it into multiple targets.

se7entyse7en:
Highly agree with @dpordomingo.

lwsanty:
Done

PR #14, opened 2019-12-18
Adds tabs logic on mini-timeseries

se7entyse7en:
Closes [ENG-65](https://athenianco.atlassian.net/browse/ENG-65?atlOrigin=eyJpIjoiMzYxMTY0ODIwZjhlNGExMTk0OTQ5Y2RkYmQ3NzE3Y2YiLCJwIjoiaiJ9).

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,16 @@
 
 Webapp for the Athenian product
 
+# Development
+
+To run a development version locally run the followings:
+```
+# install the dependencies
+$ npm install
+
+# run the development server
+$ npm start
+```
 
 # Contributing
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,15461 @@
+{
+  "name": "athenian-webapp",
+  "version": "0.1.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@babel/code-frame": {
+      "version": "7.5.5",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.5.5.tgz",
+      "integrity": "sha512-27d4lZoomVyo51VegxI20xZPuSHusqbQag/ztrBC7wegWoQ1nLREPVSKSW8byhTlzTKyNE4ifaTA6lCp7JjpFw==",
+      "requires": {
+        "@babel/highlight": "^7.0.0"
+      }
+    },
+    "@babel/core": {
+      "version": "7.7.4",
+      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.7.4.tgz",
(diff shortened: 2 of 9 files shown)


src/App.js
```diff
@@ -25,33 +24,62 @@ const addNew = data => data.map((_, i) => {
...
+];
+
+
 export default function() {
   const [pipelineData] = useState(getData());
   const [data, setData] = useState(getData());
 
+  pipeline.map(step => {
+    step.tab.data = pipelineData;
+    return;
```
kuba--: ./src/App.js
Line 67:5:  Arrow function expected a return value  array-callback-return

src/App.js
```diff
@@ -25,33 +24,62 @@ const addNew = data => data.map((_, i) => {
...
+  });
+
+  sampleCharts.map(card => {
+    card.data = data;
+    return;
```
kuba--: Line 72:5:  Arrow function expected a return value  array-callback-return

README.md
```diff
@@ -2,6 +2,16 @@
 
 Webapp for the Athenian product
 
+# Development
+
+To run a development version locally run the followings:
```
dpordomingo: Development instructions are already available in the [`Contributing`](https://github.com/athenianco/athenian-webapp/blob/master/docs/CONTRIBUTING.md) guide; we could maybe make it more explicit in the `README`, but I think it's better if we put there these instructions, and we reserve the `README` for docs about its usage.

On the other hand, we're already using `Yarn` [[1]](https://www.ryadel.com/en/yarn-vs-npm-pnpm-2019/)

package-lock.json
```diff
@@ -0,0 +1,15461 @@
+{
```
dpordomingo: We should add `package-lock.json` into the `.gitignore` to avoid conflicts with people using `npm`

src/App.js
dpordomingo: I think you wanted to use a `forEach` instead of a `map`, which does not use `return`.

Also, modifying the `pipeline` and `sampleCharts` global variables is not that funny, and does not let you see that good the tab effect (changing the stage body as you click in different tabs).

The effect will be noticeable if you replace the whole content from `line:65` to `line:84` with:
```javascript
  const data = pipeline.map((stage, i) => {
    return {
      tab: {...stage.tab, data: getData()},
      body: {charts: stage.body.charts.map(c => {
        return {...c, title:c.title+i, data: getData()}
      }) }
    }
  });
```

that will create a new `data` constant with the proper data to be passed into `Pipeline` at `line:87`
```javascipt
<Pipeline pipeline={ data } />
```

src/App.js
dpordomingo: _ditto_

dpordomingo:
I pushed 1e89a43 adding the tab effect, and addressing @kuba-- comments with [my proposal](https://github.com/athenianco/athenian-webapp/pull/14#discussion_r359694702).
I also merging this PR in order to put this into production and to keep working in [ENG-78], where I were also doing most of the changes that were also done by this PR.

README.md
se7entyse7en: yup sorry, it's the habit 😅

PR #19, opened 2020-01-08
Make the code more React-like, and update .gitignore 

dpordomingo:
per commits:

826c4f1 Make the code more React-like
- Use stateless function components, instead of clas ones
https://reactjs.org/blog/2015/10/07/react-v0.14.html#stateless-function-components
- Destructuring props in stateless function arguments
- Avoid defining HTML elements in different parts of the render, and do it at once instead

89c3b75 Add development instructions in main README.md

And other minor changes:
- b299137 Ignore `package-lock.json` from npm
- 95e893e Ignore `.ci` dir
- 1adcc78 Update dependencies


se7entyse7en:
I didn't review yet, but why removed `package-lock.json`? It should be version controlled as well.

dpordomingo:
We're handling our dependencies with `yarn`, which locks them with [`/yarn.lock`](https://github.com/athenianco/athenian-webapp/blob/master/yarn.lock). Using both lock files might be error-prone.

se7entyse7en:
Oh sorry, right. Thanks! I thought that we aren't using any lock anymore. 👍 I'll review now then.

src/components/charts/CleanAreaChart.js
```diff
@@ -2,99 +2,66 @@ import React from 'react';
...
+  var fillColor, strokeColor;
 
-    return (
-      <linearGradient id={ ns + '-' + key }
-                      x1={ gpc.x1 } x2={ gpc.x2 }
-                      y1={ gpc.y1 } y2={ gpc.y2 }>
-        { stops }
-      </linearGradient>
-    );
+  if (fill) {
```
se7entyse7en: Is this correct? The semantic changed, what if I pass a string color? Before iirc you can pass an object or a string.

src/components/charts/CleanAreaChart.js
```diff
@@ -2,99 +2,66 @@ import React from 'react';
...
-    return gradientPointsCoords;
-  }
-
-  buildGradientStops = (stops) => {
-    return stops.map(s => (
-      <stop offset={ s.offset }
-            stopColor={ hexToRGB(s.stopColor) }
-            stopOpacity={s.stopOpacity }/>
-    ));
+  if (stroke) {
```
se7entyse7en: Same here.

src/components/charts/CleanAreaChart.js
dpordomingo: That's true. I'll review this in-depth and ping you.

se7entyse7en: Thanks, I think that you can just copy the same condition and that's fine. 👍

dpordomingo: I pushed e6ca3b0cf63ab26dd46dbb4a4939eaf49cac86d1 to restore that configuration; what is really needed, is to confirm that `stops` key is defined (we could add more validations, but it could be too much at this early stage, wdyt?)

se7entyse7en: I think that just checking whether it's an object instead of a specific key would be much extensible, but not a blocker.

se7entyse7en: Ah no sorry ok, even before we were accepting only those two params. LGTM then 👍 thanks!

PR #20, opened 2020-01-09
Override bootstrap grid sizes to fit athenian layout

dpordomingo:
blocked by #19

_Changes to be reviewed only by 1127c522bf056a2798c4398c1ee1a4550370855b_

Override bootstrap grid sizes to fit athenian layout:

Main new values:
- grid-gutter-width: 20px instead of 30px
- grid-breakpoints: xl -> 1300px instead of 1200px
- container-max-widths: xl -> 1270px instead of 1140px

see original values at https://getbootstrap.com/docs/4.3/layout/grid/#variables

### DIFF
--- a/src/sass/main.scss
+++ b/src/sass/main.scss
@@ -1,4 +1,4 @@
-$body-bg: #f5f5f5;
+@import './overrides.scss';
 
 @import '~react-vis/dist/style.css';
 
--- a/src/sass/overrides.scss
+++ b/src/sass/overrides.scss
@@ -1,0 +1,33 @@
+/**
+ * Override grid sizes
+ * https://getbootstrap.com/docs/4.3/layout/grid/#variables
+ */
+
+$grid-gutter-width: 20px;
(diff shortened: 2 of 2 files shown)


se7entyse7en:
Could you please for these kinds of changes attach a screenshot? It would help as it would avoid me to locally clone and run 🙏

se7entyse7en:
BTW don't focus too much ad this point about the spacing, the layout of the whole page, borders, etc. as things will change, and potentially a lot. Let's focus on the agreed pieces with @warenlg and we'll then do a final styling review.

dpordomingo:
I updated the PR desc with two screenshots showing that the `.container` is now `1240px` width as in the design.
Also, we can now also adapt the grid on our will, so as soon as Product delivers new specs for the layout, we can proceed with the change using the new variables.

These changes were really needed in order to make the pipeline thumbnails wider, so there will fit all the info that Product requested.

PR #21, opened 2020-01-09
ENG-81 Complete the pipeline thumbnails

dpordomingo:
fix [[ENG-81]] Finalize the cards of the pipeline by adding the extra information

## Changes:
- 3a5384f45c90fac8d5a86c1c5ba7d279acd94320 Completes the pipeline thumbnails

## Requirements to achieve the above:
- 0ed2f4b Fine tunning of the layout to fit the design, and create space in the thumbnails; And also some sass assets rearrangements.
- f8256b3 Use a fork of **SB Admin 2** which enables grid gutter config (see details at #22)

image

[ENG-81]: https://athenianco.atlassian.net/browse/ENG-81

### DIFF
--- a/src/components/charts/CleanAreaChart.js
+++ b/src/components/charts/CleanAreaChart.js
@@ -1,10 +1,10 @@
 import React from 'react';
 
-import { XYPlot, AreaSeries, GradientDefs } from 'react-vis';
+import { FlexibleXYPlot, AreaSeries, LineSeries, GradientDefs } from 'react-vis';
 
 import { hexToRGBA } from '../../utils/colors';
 
-export default ({ fill, stroke, width, height, data }) => {
+export default ({ fill, stroke, data }) => {
 
   const key = () => Math.random().toString(36).substring(2, 15);
   const gradients = [];
@@ -21,22 +21,20 @@ export default ({ fill, stroke, width, height, data }) => {
 
   if (stroke && stroke.stops) {
     const id = 'StrokeGradient-' + key();
-    gradients.push(<LinearGradient key={id} config={stroke} />);
+    gradients.push(<LinearGradient key={id} id={id} config={stroke} />);
     strokeColor = 'url(#' + id + ')';
   } else {
     strokeColor = stroke;
   }
 
   return (
-    <XYPlot width={width} height={height}>
+    <FlexibleXYPlot>
       <GradientDefs>
         {gradients}
       </GradientDefs>
-      <AreaSeries data={data}
-        fill={fillColor}
-        stroke={strokeColor}
-      />
-    </XYPlot>
+      <LineSeries data={data} color={strokeColor} animation="stiff" />
+      <AreaSeries data={data} stroke="none" fill={fillColor} animation="stiff" />
+    </FlexibleXYPlot>
   );
 };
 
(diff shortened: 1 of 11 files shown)


se7entyse7en:
I'm reviewing and adding @warenlg as a reviewer as well. Thanks for the screenshots! I guess that would make especially @warenlg's review much easier so that he can focus on the output instead of the code without checking out and running the server 👍

src/components/charts/CleanAreaChart.js
```diff
@@ -21,22 +21,20 @@ export default ({ fill, stroke, width, height, data }) => {
...
+    <FlexibleXYPlot>
       <GradientDefs>
         {gradients}
       </GradientDefs>
-      <AreaSeries data={data}
-        fill={fillColor}
-        stroke={strokeColor}
-      />
-    </XYPlot>
+      <LineSeries data={data} color={strokeColor} animation="stiff" />
```
se7entyse7en: So now the "top border" of the area chart is actually a `LineSeries` and the "background" is the `AreaSeries`, right? If so seems much more logical 👍

src/components/pipeline/PipelineCard.js
```diff
@@ -1,29 +1,48 @@
...
+
+  return (
+    <div className={"card shadow pipeline-thumbnail" + (active ? ' active' : '')} onClick={onClick} style={cardStyle}>
+      <div className="card-body">
+        <div className="row no-gutters card-title text-xs text-uppercase">
+          <div className="col font-weight-bold">{title}</div>
+          <div className="col-auto">
+            <span className="badge badge-pill badge-secondary align-middle py-1 px-2">{badge}</span>
+          </div>
+        </div>
```
se7entyse7en: Couldn't this go in a `card-head` or `card-header` (I guess there's something similar)? In the end, it's really a header. And also the way you structured it is separated from the chart.

src/components/charts/CleanAreaChart.js
dpordomingo: That's true :D

src/components/pipeline/PipelineCard.js
dpordomingo: Using `card-header` and `card-body` renders this kind of cards

https://startbootstrap.com/previews/sb-admin-2/

Using `card-title` inside `card-body` creates this other 

https://getbootstrap.com/docs/4.3/components/card/#titles-text-and-links

pros of the second one:
- closer to our design using plain Bootstrap and SB Admin 2
- avoid us to start overriding the header to remove unnecessary borders, background...

dpordomingo: About your last phrase:
>_And also the way you structured it is separated from the chart._

What do you mean?

dpordomingo:
Should I wait for @warenlg approval? I'd say no; we can deploy into prod as soon as you @se7entyse7en approves this, and let him review the changes in Prod, and open issue or add comments here if there is any other change needed.

src/components/pipeline/PipelineCard.js
se7entyse7en: My comment was only related to a "logical/structure" point of view. And by looking at your code you already separated into two different divs the part related to the chart and the part related to the header, that's why I thought that maybe it would have made sense to use `card-title`. But since it's more related to styling let's not overthink it 👍

dpordomingo: with bootstrap semantic separation between elements and styles is difficult... sometimes impossible if you don't want to rewrite all its styles.

PR #22, opened 2020-01-09
[RFC] Use a fork of SB Admin 2 which enables grid gutter config

dpordomingo:
depends on #21 

Changes are only:

1f38243 Use `@dpordomingo/startbootstrap-sb-admin-2` which lets us to override bootstrap gutter and all other bootstrap variables
>In order to reduce the grid gutter overriding bootstrap value it is needed to fix BlackrockDigital/startbootstrap-sb-admin-2#320
>
>Untill the fix (BlackrockDigital/startbootstrap-sb-admin-2#321) is merged and released, it will be used [`@dpordomingo/startbootstrap-sb-admin-2 v4.0.8`](https://github.com/dpordomingo/startbootstrap-sb-admin-2/tree/v4.0.8) which already fixes the described issue

abdac8f Revert "HACK. Increase page width instead of reduce the gutter"
>No longer needed thanks to `@dpordomingo/startbootstrap-sb-admin-2 v4.0.8`


dpordomingo:
Since this is approved, I'll add it to https://github.com/athenianco/athenian-webapp/pull/21 in order to be merged into `master` at once.

PR #23, opened 2020-01-10
ENG-83 Use API metrics from thumbnails

dpordomingo:
fix [[ENG-83]] Call the API to fill the cards with the data returned

This PR calls the API to get thumbnails metrics. It's still a _poc_ to be evolved as we start adding use cases and have real data and stop mocking.

## TODO for the API
The following improvements in the API side would be very welcome in the frontend:

- [[ENG-122]] Only zero-values in `lead-time`, `wait-first-review-time`, `merging-time` and `release-time`

image
Some thumbnails lack chart because per [[ENG-122]] its metrics returns only `zero-values`.

[ENG-83]: https://athenianco.atlassian.net/browse/ENG-83
[ENG-125]: https://athenianco.atlassian.net/browse/ENG-125
[ENG-122]: https://athenianco.atlassian.net/browse/ENG-122

### DIFF
--- a/src/services/api.js
+++ b/src/services/api.js
@@ -1,0 +1,144 @@
+export const getPipelineDataInitial = () => getPipelineData([], () => []);
+
+const SECOND = 1;
+const MINUTE = 60 * SECOND;
+const HOUR = 60 * MINUTE;
+
+const getHours = secondsString => {
+  if (!secondsString) {
+    return 0;
+  };
+
+  const seconds = parseFloat(secondsString);
+  if (!seconds) {
+    return 0;
+  };
+
+  return seconds / HOUR;
+};
+
+export const getPipelineDataAPI = () => {
+  const metrics = ['pr-lead-time', 'pr-wip-time', 'pr-wait-first-review-time', 'pr-review-time', 'pr-merging-time', 'pr-release-time'];
+
+  return api(metrics).then(apiData => {
+    const thumbsData = [];
+
+    apiData.calculated[0].values.forEach(day => {
+      day.values.forEach((value, metricNo) => {
+        if (!thumbsData[metricNo]) {
+          thumbsData[metricNo] = [];
+        }
+
+        thumbsData[metricNo].push({ x: new Date(day.date), y: getHours(value) });
+      });
+    });
+
+    return getPipelineData(thumbsData, getRandData);
+  })
+};
+
+const sampleCharts = [
+  {
+    title: 'Time to Commit in Base Branch',
+    color: '#62C2DF',
+    insights: [
+      { title: "insight 1", value: "foobar 1" },
+      { title: "insight 2", value: "foobar 2" }
+    ]
+  },
+  {
+    title: 'Time to Release',
+    color: '#FA1D62',
+    insights: [
+      { title: "insight 3", value: "foobar 3" }
+    ]
+  },
+  {
+    title: 'Merged Pull Request',
+    color: '#FF7D3A',
+    insights: []
+  },
+];
+
+const pipeline = [
+  {
+    tab: { title: 'PR created', color: '#FF7D3A', text: '2 days', badge: 10, metric: 'pr-lead-time' },
+    body: { charts: sampleCharts }
+  },
+  {
+    tab: { title: 'Work in progress', color: '#FA1D62', text: '5 weeks', badge: 235, metric: 'pr-wip-time' },
+    body: { charts: sampleCharts }
+  },
+  {
+    tab: { title: 'first review', color: '#FF7D3A', text: '10 months', badge: 5, metric: 'pr-wait-first-review-time' },
+    body: { charts: sampleCharts }
+  },
+  {
+    tab: { title: 'approval', color: '#9460DA', text: '4 days', badge: 15, metric: 'pr-review-time' },
+    body: { charts: sampleCharts }
+  },
+  {
+    tab: { title: 'PR merged', color: '#FF7D3A', text: '5 weeks', badge: 33, metric: 'pr-merging-time' },
+    body: { charts: sampleCharts }
+  },
+  {
+    tab: { title: 'release', color: '#62C2DF', text: '3 days', badge: 12, metric: 'pr-release-time' },
+    body: { charts: sampleCharts }
+  },
+];
+
+const api = metrics => {
+  const repos = [
+    'github.com/athenianco/athenian-webapp',
+    'github.com/athenianco/athenian-api',
+    'github.com/athenianco/metadata',
+    'https://github.com/athenianco/metadata-retrieval'
+  ];
+  const granularity = 'week';
+
+  const dateFrom = '2019-11-19';
+  const dateTo = '2020-02-11';
+
+  const query = {
+    'for': [{
+      'repositories': repos,
+    }],
+    'metrics': metrics,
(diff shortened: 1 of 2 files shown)


dpordomingo:
CORS no longer blocks this PR

se7entyse7en:
Tests failing

dpordomingo:
doh!
facebook/create-react-app build fails when building in CI if there is any warning!
https://github.com/facebook/create-react-app/issues/2453#issuecomment-306892544

dpordomingo:
[facebook/create-react-app](https://github.com/facebook/create-react-app) build fails when building in CI if there is any warning, and [they do recommend to unset `CI` env var](https://github.com/facebook/create-react-app/issues/2453#issuecomment-306892544) in such cases!
And no way :no_good_man: 
So warnings were removed, and build now passes.

src/services/api.js
```diff
@@ -0,0 +1,153 @@
...
+
+  const options = {
+    method: 'POST',
+    body: JSON.stringify(query)
+  };
+
+  let url = 'http://34.96.127.30/v1/metrics_line';
+
+  return fetch(url, options)
+    .then(response => response.json());
```
se7entyse7en: We should add handling of non-200 responses.

src/services/api.js
```diff
@@ -0,0 +1,153 @@
...
+    'date_to': dateTo,
+    'granularity': granularity,
+  };
+
+  const options = {
+    method: 'POST',
+    body: JSON.stringify(query)
+  };
+
+  let url = 'http://34.96.127.30/v1/metrics_line';
```
se7entyse7en: Let's add a `config.js` in `public` with the following content:
```
window.ENV = {
    API_BASE_URL: http://34.96.127.30
}
```
so that we can write:
`let url = window.ENV.API_BASE_URL + '/v1/metrics_line';`

src/services/api.js
```diff
@@ -0,0 +1,153 @@
+export const getPipelineDataInitial = () => getPipelineData([], () => []);
```
se7entyse7en: Is this to avoid blocking the first `render`?

dpordomingo: nope. `getPipelineDataInitial` returns the initial data (empty values) for the pipeline data, that will be used to initialize the `pipelineState` React state.

### explanation:
Before we passed the data to the pipeline as a known object called `data`, now we pass it as React a state called `pipelineState` ([here is the change](https://github.com/athenianco/athenian-webapp/pull/23/files/015eb8f224a9f2f79b1b5d590933ffa4c4bda516#diff-14b1e33d5bf5649597cdc0e4f684daddR46))

`pipelineState` state is created and initialized by `useState` at [`src/App.js#R10`](https://github.com/athenianco/athenian-webapp/pull/23/files/015eb8f224a9f2f79b1b5d590933ffa4c4bda516#diff-14b1e33d5bf5649597cdc0e4f684daddR10), and its initial value comes from `getPipelineDataInitial`
```js
```

When the component is rendered, `useEffect` is in charge of assigning the proper value to the `pipelineState` at [`src/App.js#R12`](https://github.com/athenianco/athenian-webapp/pull/23/files/015eb8f224a9f2f79b1b5d590933ffa4c4bda516#diff-14b1e33d5bf5649597cdc0e4f684daddR12) thanks to `setPipelineData` from above.
```js
useEffect(() => {
    getPipelineDataAPI().then(data => {
        setPipelineData(data);
    });
}, []);
```

More info in react docs about [`state`](https://reactjs.org/docs/hooks-state.html) and [`effect`](https://reactjs.org/docs/hooks-effect.html) hooks.

src/services/api.js
dpordomingo: I'm planning to use open API generator and its way to define the api, so this could maybe handled by its generator. This might be also the way, but I'd prefer to handle this separatelly and as part of the other task.

src/services/api.js
dpordomingo: This would be handled by API client.

dpordomingo:
I planned to use [OpenAPI generator](https://openapi-generator.tech/#try) to generate the javascript API client from the [webapp specs](https://raw.githubusercontent.com/athenianco/athenian-api/master/server/athenian/api/openapi/openapi.yaml), so many things would be handled by the autogenerated client, like the error handling and or the URL API.

As I said in the PR desc, this should be only a POC to confirm that the API and the frontend can work in production, and there is no problem with CORS, certs and such.
If you don't oppose that plan, I'd like progress in that direction.

**PS.** I just tried the generator, and it seems to work, so I'm planning how to add it to the project, but it's something that, as I said, I'd like to consider separately.

src/services/api.js
se7entyse7en: Okay, let's see then how this part will be filled by the generator. The way of using this kind of configuration using a file in `public` folder (and adding it as with a `script` tag in the `index.html`) makes it easy to be configured using a k8s `configMap`.

src/services/api.js
se7entyse7en: yup, what I meant is why you don't populate with real data since the beginning instead of initializing with empty values? I mean AFAIK there's no constraint in initializing the state through `useState` with any data. That's why I was asking whether you did it to avoid slowing down the first render.

se7entyse7en:
Regarding using the auto-generate client I didn't know that you were planning to use it, but it's a good idea 👍. 

Regarding certs I'm pretty sure that it won't work as it's not possible to call http from https, but we're working on it.

src/services/api.js
se7entyse7en: Ok, I didn't know you were planning to use the auto-gen client 👍

src/services/api.js
dpordomingo: Not really (if I understood you).

The data needed for the pipeline is stored under `pipelineState` ReactState.
The problem is that to fetch it, we call the API, so it's not yet ready when rendering the React component.

React, by design, [can not handle promises in their props](https://github.com/facebook/react/issues/6481#issuecomment-209606113) nor children, so you need to put there real data. Since there is still no data when the component is being rendered, it is used `zero-data`.

The pattern is described here as [fetch-on-render](https://reactjs.org/docs/concurrent-mode-suspense.html#approach-1-fetch-on-render-not-using-suspense).

We will iterate more on fetching and rendering once we have more metrics and api calls, trying to improve the loading and UX when there is no data; but at this point, we will also need to define how the app loads and how should it behave on missing or incomplete data.

**addendum**. React mates are working in [Suspense api](https://reactjs.org/docs/concurrent-mode-suspense.html), which would "_ let your components “wait” for something before they can render_" but it is not yet ready.

dpordomingo:
I answered your two pending questions.

If we merge this PR, prod will use the current API, and it will fail if the config is not properly done.
Do you think that we should:
a) wait, or
b) merge/deploy, and then see how it fails?

I'd say `b)` because it could help @lwsanty to know what to fix, and also it would help him to see if his solution worked.

But we can do `a)`, blocking this PR till you both think we can merge.

se7entyse7en:
Let's merge as CORS has been fixed and also Certs has been added, so we shouldn't have any problem. Let's touch again the topic on setting the api domain on a config file once you've defined the structure with generated client.

dpordomingo:
The API is now broken, https://athenianco.atlassian.net/browse/ENG-138
I'll merge once that is fixed.

PR #26, opened 2020-01-17
Add registration/login flow and logout

se7entyse7en:
This PR closes [ENG-145](https://athenianco.atlassian.net/browse/ENG-145). The main goal of this PR is to add the integration with Auth0 SDK. Being authenticated is not a requirement for accessing the webapp for now.

This adds:
- a `/login` path that shows a dummy page tbd with just a login button that when clicked redirects to auth0 authentication page,
- a `/logout` path that only serves for testing `/login` multiple times.

If already authenticated, then `/login` simply redirects to the home.
If not already authenticated, then `/logout` will just print an error.

Here's the screenshot of the dummy page:

Screenshot 2020-01-17 at 13 57 47

In order to be able to test it, `config.js` must have the `auth` key with `domain` and `clientId` set according to application settings in auth0:

```
window.ENV = {
  auth: {
    domain: "<auth0 domain>",
    clientId: "<auth0 client id>",
  }
};
```

I tested this with a personal account on Auth0 with GH SSO enabled and a GH Oauth application linked to Auth0.


se7entyse7en:
For now, if merged it won't work as it requires update on deployment.

se7entyse7en:
@lwsanty as discussed on Slack I should be able to change the infrastructure part as well 👍, I'll ping you in case my mind blows out 😂

public/index.html
```diff
@@ -15,6 +15,7 @@
       user's mobile device or desktop. See https://developers.google.com/web/fundamentals/web-app-manifest/
     -->
     <link rel="manifest" href="%PUBLIC_URL%/manifest.json" />
+    <script type="text/javascript" src="%PUBLIC_URL%/config.js"></script>
```
dpordomingo: Why not using regular `process.env` to define configurable values? It would also let us configure them easily through the standard `.env` files and passing env values to the `make` and `yarn` commands. You can see an example at 447e85902b0f7d6f0ae44feca93c2ce8c189174e

src/App.js
```diff
@@ -1,59 +1,34 @@
...
-        <div className="container my-auto">
-          <div className="copyright text-center my-auto">
-            <span>Copyright &copy; Athenian.co 2020</span>
-          </div>
-        </div>
-      </footer>
-
+      <Router history={history}>
+        <Switch>
+          <Route path="/login">
```
dpordomingo: I could configure the app with a personal Auth0 account, but I'm not sure if I did it properly. If you access the home without being logged in, it works; you can access `/login` every time you want, no matter if you already logged in. Is that correct, or did I missed any configuration? If I was right, I would expect that:
  - accessing `/login` once you logged in, would automatically redirect you to the home,
  - it is needed to open `/logout` in order to be able to use `/login` again.

If you want to address this later, it lgtm.

src/react-auth0-spa.js
```diff
@@ -0,0 +1,86 @@
+import React, { useState, useEffect, useContext } from "react";
```
dpordomingo: Could you please add simple docs in `docs/CONTRIBUTING.md` about how to configure it once we have an agreement on how to define its config (using `config.js` or plain env vars)?

It's not needed a lot of details, maybe it only needed to explain that it's needed an Auth0 account, its minimum config (I think it's only needed to define the same `http://localhost:3000` in `Allowed Callback URLs`, `Allowed Web Origins` and `Allowed Logout URLs`; and how to define Auth0 config params in the development env.

se7entyse7en:
> I discovered this PR is based on https://auth0.com/docs/quickstart/spa/react/01-login, being mostly a copy-paste of Auth0 getting started guide. Knowing it in advance would have speed up the review when I first realized about this PR.

You're completely right, I'm really sorry this is totally my fault :(

public/index.html
se7entyse7en: I'll comment directly in the PR 👍

src/App.js
se7entyse7en: yup, this is the expected from this PR. Since user flow wasn't defined yet, I just added this page that makes you login. Then I added a logout because otherwise as you noticed as well you cannot login again (or better you cannot test again the login page). The behaviour will be better implemented in following prs.

dpordomingo: But I did login many times without visiting `/logout`. In the final version it would be a big. But is it expected at this stage, or did I do something wrong when testing?

se7entyse7en: How are you testing? I've just retested locally and if I go multiple times to `/login` it redirects me to home page as expected.

dpordomingo: In my case it only redirects me to the home on Auth0 callback, but if I navigate later to `/login` it does not redirect me and let's me to login again against Auth0.

I think I missed some config, I'll review it once you post the docs.

src/index.js
```diff
@@ -6,7 +6,30 @@ import App from './App';
 
 import * as serviceWorker from './serviceWorker';
 
-ReactDOM.render(<App />, document.getElementById('root'));
+import { Auth0Provider } from "./react-auth0-spa";
```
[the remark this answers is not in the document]
se7entyse7en: done [here](https://github.com/athenianco/athenian-webapp/pull/26/commits/2e115aa3e94241d07dfb69cb8a0e44a49bea1d69).

src/pages/Home.js
```diff
@@ -0,0 +1,56 @@
...
+export default () => {
+  const [pipelineState, setPipelineData] = useState(getPipelineDataInitial());
+
+  useEffect(() => {
+    getPipelineDataAPI().then(data => {
+      setPipelineData(data);
+    });
+  }, []);
+
+  return [
```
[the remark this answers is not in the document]
se7entyse7en: done [here](https://github.com/athenianco/athenian-webapp/pull/26/commits/2f4728226aa7a92718b98debe5f86cc5d24bad5e) using an empty wrapper tag in order to avoid adding an unneeded div tag.

src/react-auth0-spa.js
se7entyse7en: done [here](https://github.com/athenianco/athenian-webapp/pull/26/commits/d4665f1a7cc70acd23015361fc993699104da6e5).

src/App.js
se7entyse7en: @dennwc can we create a testing application on auth0 configure exactly as the production one just with the required urls pointing to `http://localhost:3000`?

se7entyse7en: @dpordomingo probably the authorization callback on GH is not set properly? So the login actually fails?

src/pages/Home.js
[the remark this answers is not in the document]
dpordomingo: wow! it's so dirty from the HTMLx point of view... and so funny workaround from the React mates... that I think I love it. Thanks for finding it!

src/react-auth0-spa.js
dpordomingo: Many thanks!!!

Trying to follow your guide, I had to do the things in a slightly different order, so I'd document it the following way, wdyt?

```markdown
### Auth0 and GitHub App local testing

For locally testing the authentication, you need to configure Auth0 and GitHub
to work together. Follow these steps or refer to Auth0 and GitHub docs.

#### 1. Register Single-Page Applications in Auth0 ([docs](https://auth0.com/docs/dashboard/guides/applications/register-app-spa))

1. Create an Auth0 account,
2. Create an application with "Application Type" set to "Single Page Application",
using the following settings (eventually change the port with the one you plan to use):
    - "Allowed Callback URLs": `http://localhost:3000`
    - "Allowed Web Origins": `http://localhost:3000`
    - "Allowed Logout URLs": `http://localhost:3000/logout`

#### 2. Connect your Auth0 app to a GitHub OAuth App ([docs](https://auth0.com/docs/connections/social/github))

1. Register a new OAuth App in GitHub,  with these settings:
    - "Homepage URL": `http://localhost:3000`
    - "Authorization callback URL": `https://<auth0 domain>/login/callback`
    (e.g., `your-app.eu.auth0.com`)
2. Enable the connection with GitHub from your Auth0 Dashboard providing the
`Client ID` and `Client Secret` of the GitHub OAuth App from the previous step, and ask for these permissions:
    - `repo:user`
    - _¿any other?_

#### 3. Configure Athenian App

Update the [`config.js`](https://github.com/athenianco/athenian-webapp/blob/master/public/config.js)
file by setting both `auth.domain` and `auth.clientId` with the corresponding values of the Auth0 application.
```

src/App.js
dpordomingo: Wow!
Connecting with GH! I totally missed that!
I could finally reproduce it following https://github.com/athenianco/athenian-webapp/pull/26#discussion_r369951030

public/index.html
dpordomingo: let's take the `conf.js` approach as decided in #31

se7entyse7en:
Updated documentation as you suggested @dpordomingo, merging once CI is green 👍.

PR #30, opened 2020-01-22
ENG-134 Create DoubleYAxis chart

dpordomingo:
solves [[ENG-134]]

This kind of parallel-axis charts are not `react-vis` built-in, so it was needed to join some of its components and to massage some data. (see how they implement [another example](https://github.com/uber/react-vis/blob/master/showcase/axes/parallel-coordinates-example.js))

image

current limitations:
- The `DecorativeAxis` from `react-vis` has an issue that does not let you to properly configure its color, stroke, number of ticks, nor title.

[ENG-134]: https://athenianco.atlassian.net/browse/ENG-134


se7entyse7en:
Seems a lot complicated, unfortunately, and I read the issues on `react-viz` repo as well. :(
I'm not reviewing the part specific to `react-vis` since it works. Could you please just confirm that the y-axis label on the right is shown? As from the screenshot I can only see the left one.

BTW just as a reminder for the future, moreover at this stage where product requirements can be highly discussed, please raised the issue if a chart is not built-in or requires a lot of hand-made work. In that case, we can discuss with @warenlg whether we could avoid addressing it by simply changing the chart. Of course, if it's a strong product requirement then we'll tackle it, but if we can avoid it the better.

dpordomingo:
What you ask about the right `y-axys` label, is directly affected by the issue I mentioned about the `DecorativeAxis` in the PR desc.
>_current limitations:_
>_- The `DecorativeAxis` from `react-vis` has an issue that does not let you to properly configure its color, stroke, number of ticks, nor title._

Anyhow, I didn't want to spend much more time debugging the issue, nor developing a workaround; I just preferred to release something close to the design, and ready to be used once we have data.
There could be alternatives, but I think it could be more rewarding to invest that time in the whole dashboard.

se7entyse7en:
> What you ask about the right y-axys label, is directly affected by the issue I mentioned about the DecorativeAxis in the PR desc.

I got it, but it's not about the configuration, in the screenshot it doesn't appear at all. There should be I guess a "PR created" label on the right similarly to the "WIP time, hours" there's on the left. If we cannot have it, I strongly think that we cannot ship this like this (but the last word on this is on you @warenlg). Maybe it's on the right and just cropped in the screenshot?

> Anyhow, I didn't want to spend much more time debugging the issue, nor developing a workaround; I just preferred to release something close to the design, and ready to be used once we have data.

What do you mean? IMO this is a workaround since it's not cleanly supported by `react-viz`. But again, if it works (also has the right y-axis label) for me it's not a problem to merge it since we already have it.

> There could be alternatives, but I think it could be more rewarding to invest that time in the whole dashboard.

Sorry, I don't understand this. This is what I'm saying that at this stage it would be better to discuss these kinds of obstacles also with @warenlg to understand whether an easier and faster alternative could be found that still fulfills the requirements.

warenlg:
Don't waste time too much time with this kind of graph. We can show the metrics in another way, I did not know it was such a pain in the ass to implement. Please start with the easiest charts to compile.

se7entyse7en:
@dpordomingo l'm closing this, since we have updated charts from product.

PR #31, opened 2020-01-22
[RFC] Use env to pass config values

dpordomingo:
caused by https://github.com/athenianco/athenian-webapp/pull/26#discussion_r369438437

If this is accepted, the configuration will be done using `.env` files, or overrided with environment variables.

For example:
Running the following will override Auth0 conf and change the API URL
```
REACT_APP_API_URL=http://my_custom_api REACT_APP_AUTH0_DOMAIN=custom.auth0.com REACT_APP_AUTH0_CLIENT_ID=XXXX yarn start

```

It could be done the same creating `.env.development` in the app root folder with this content:
```env
REACT_APP_API_URL = http://my_custom_api
REACT_APP_AUTH0_DOMAIN = custom.auth0.com
REACT_APP_AUTH0_CLIENT_ID = XXXX
```


dpordomingo:
From my understanding of the `conf.js` approach, the problem would be the same: how to make dynamic an static page built inside a docker image.

For the most common usecase: generating the proper image for the proper environment, `.env` approach is the most straightforward bec it is interpreted on build time by yarn compiler, introducing into the image directly the proper value from its proper `.env`.

If we want to change the config using the very same image, we should take `conf.js` approach and then pass the proper `conf.js` file to the container, mounting it as a volume when running the container.

Wdyt?
/cc @lwsanty

se7entyse7en:
That's why I wanted to use the `conf.js` approach, because it let us have a single image independently from the environment and the settings are passed at runtime. With k8s it's easy to setup as it's only a matter of defining a configmap and mount it as a volume as you said.

dpordomingo:
Closed in benefit of `conf.js` approach

PR #33, opened 2020-01-23
[Hotfix] Update node-alpine base image

dpordomingo:
There was a problem rebuilding node-gyp which also seems to need python2 and is not available. It's fixed in the latest release of `node-alpine`.
I guess the error appeared when trying to build any new dependency added (or updated) by https://github.com/athenianco/athenian-webapp/pull/25

Also it was needed the  inside the image in order run 'make' commands to install and build the app

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,16 @@
 # Build web app
 # ----------------
-FROM node:13.1-alpine AS app-builder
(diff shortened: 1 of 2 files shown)


dpordomingo:
I merged bec it's blocking the deploy of the app, but mentioning @lwsanty in case he wants to review this.

PR #32, opened 2020-01-23
Add /waiting endpoint with carousel

se7entyse7en:
Closes: https://athenianco.atlassian.net/browse/ENG-149

This is the content of the `/waiting` page with the scaffolding for adding more proper content once defined:

Screenshot 2020-01-23 at 14 25 29

Changes included in this PR:
1. added a protected page at `/waiting` with content placeholders (`PrivateRoute` component copied from Auth0 quickstart),
2. the waiting page redirects to home after a fixed amount of time for now,
3. made the `logout` button in the user dropdown work by calling the `/logout` endpoint,
4. changed the authorization callback to `/callback` (it will be useful also for future PRs),
5. redirect to waiting page after successful login,
6. added some components for reusing the layout of a logged-in page.

PR #34, opened 2020-01-23
Improve auth UX

dpordomingo:
caused by #32

Some minor improvements for the UX:
- Afther logging out, the user is redirected to the Home.
- If there is no user logged in, it won't be requested the user data to the API

### DIFF
--- a/src/js/Routes.test.js
+++ b/src/js/Routes.test.js
@@ -3,7 +3,15 @@ import { act } from 'react-dom/test-utils';
 import ReactDOM from 'react-dom';
 import Routes from 'js/Routes';
 
+import * as Auth0 from 'js/services/react-auth0-spa';
+
 it('renders without crashing', () => {
+
+  jest
+    .spyOn(Auth0, 'useAuth0')
+    .mockImplementation(() => ({}))
+
+
   const div = document.createElement('div');
   act(() => {
     ReactDOM.render(<Routes />, div);
--- a/src/js/components/layout/Navbar.jsx
+++ b/src/js/components/layout/Navbar.jsx
@@ -1,4 +1,7 @@
 import React from 'react';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import { faSignOutAlt } from '@fortawesome/free-solid-svg-icons'
+
 import logo from 'images/logos/logo-transparent.png';
 
 export default ({ user }) => (
@@ -31,7 +34,7 @@ const User = ({ user }) => user ? (
       </a>
       <div className="dropdown-menu dropdown-menu-right shadow animated--grow-in" aria-labelledby="userDropdown">
         <a className="dropdown-item" href="/logout">
-          <i className="fas fa-sign-out-alt fa-sm fa-fw mr-2 text-gray-400"></i>
+          <FontAwesomeIcon icon={faSignOutAlt} className="ml-2 text-secondary mr-2 text-gray-400" />
           Logout
         </a>
       </div>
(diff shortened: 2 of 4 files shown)


src/App.test.js
```diff
@@ -2,7 +2,15 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import App from './App';
 
+import * as Auth0 from './react-auth0-spa';
+
 it('renders without crashing', () => {
+
+  jest
+    .spyOn(Auth0, 'useAuth0')
+    .mockImplementation(() => ({}))
```
dpordomingo: This commit c73780cfed3fb5b02d6f6d71eede7ac6ea7dc807 mocks `useAuth0` hook with an empty context value.

Auth0 React wrapper (`react-auth0-spa.js`) is initializing the `Auth0Context` [without a value](https://github.com/athenianco/athenian-webapp/blob/master/src/react-auth0-spa.js) because it will be assigned later on rendering.

Turns out that tests do not behave like the browser regarding the React Context and Hooks, so, in that case, it returns nothing when called from a component rendered by the test suite.

In order to avoid a null pointer exception in those situations,
```
   "TypeError: Cannot read property 'isAuthenticated' of undefined"
```
it was needed to mock the content returned by `useAuth0` in order to get some value from there

TL;DR
I got a hint to get this solution in this post about [Testing ‘useContext’ React hook](https://medium.com/7shifts-engineering-blog/testing-usecontext-react-hook-with-enzyme-shallow-da062140fc83)

src/pages/auth/Logout.jsx
```diff
@@ -1,10 +1,19 @@
...
+  const history = useHistory();
   const { loading, isAuthenticated, logout } = useAuth0();
 
+  useEffect(() => {
+    const timer = setTimeout(() => {
+      history.push('/');
+    }, 1000);
+    return () => clearTimeout(timer);
+  }, [history]);
+
```
se7entyse7en: This is not needed, the proper way to handle this is to pass a `returnTo` option to `logout`. Have a look here:
- https://github.com/auth0/auth0-spa-js/blob/902062723784d29b1bbeac9ba634071a51e38c4b/src/global.ts#L191
- https://auth0.com/docs/logout/guides/redirect-users-after-logout
- https://github.com/auth0/auth0-spa-js/blob/ee53bcb4ddd8c2620392195dc04096046a61e47f/static/index.html#L104

dpordomingo: Maybe I misunderstood the purpose of the `/logout` page:
Should it be the final stage when the user logs out? Or should it redirect to the user after saying _goodbye_?

possibilities:
- `onClick={() => logout({ returnTo: 'http://localhost:3000/' })}` &rarr; `/logout` won't be visited.
- `onClick={() => logout({ returnTo: 'http://localhost:3000/logout' })}` &rarr; `/logout` will be final.

se7entyse7en: `/logout` is called `onClick` of the `logout` button in the dropdown. See [here](https://github.com/athenianco/athenian-webapp/blob/master/src/components/layout/Navbar.jsx#L23).

Once called this "logic" page is called in order to call the `logout` function provided by the Auth0 SDK.

We could avoid having this "page" and add the first alternative that you mentioned directly in the navbar. The reason why I separated this, is in case this permits us to possibly extend the behaviour of logout more easily. For example, the login will need to open a new tab asking for GH App installation. So for consistency, I did the same for logout.

So the dropdown button will go redirect to `/logout` and will call `logout({returnTo: 'http://localhost:3000/'})`. This will cause the logout through Auth0 that will then redirect to the home.

dpordomingo: done

PR #36, opened 2020-01-27
Complete skeleton

dpordomingo:
This PR only completes the skeleton of the site considering our design but it does not add any other special functionality which is not bootstrap built-in.

### new (small) UI elements to be used everywhere:
- dc7b7ce Badge
- 5792021 Typography
- 0615e6f Info

### mayor layout components
- e1e2a91 MainMetrics and Filters
- c41b4b6 Breadcrumbs navbar
- 8d616f3 PullRequests section

### cosmetics:
- 7efa9b8 Remove shadows

screenshot-localhost_3000-2020 01 (1)


src/js/pages/Home.jsx
```diff
@@ -1,12 +1,14 @@
...
 import Pipeline from 'js/components/pipeline/Pipeline';
+import PullRequests from 'js/components/pipeline/PullRequests';
 import Page from 'js/pages/templates/Page';
 
-import { getPipelineDataInitial, getPipelineDataAPI, getUser } from 'js/services/api';
+import { getPipelineDataInitial, getPipelineDataAPI, getPRs } from 'js/services/api';
 
 export default () => {
   const [pipelineState, setPipelineData] = useState(getPipelineDataInitial());
+  const [prsState, _] = useState(getPRs);
```
se7entyse7en: Shouldn't this be:
```suggestion
  const [prsState, _] = useState(getPRs());
```
?

src/sass/main.scss
```diff
@@ -27,3 +27,15 @@
         }
     }
 }
+
+#mainBreadcrums {
```
se7entyse7en: Typo

src/js/components/pipeline/PullRequests.jsx
```diff
@@ -0,0 +1,94 @@
...
+        };
+    }
+
+    useEffect(() => {
+        $('#dataTable').DataTable && $('#dataTable').DataTable({
+            searching: false,
+            ordering: false
+        });
+    }, [])
+
```
se7entyse7en: I'm a bit concerned about mixing JQuery and Reactjs, what are your thoughts on this? The point is that React uses the Virtual DOM while JQuery, if it need to, will change the actual DOM. I'm concerned that in some cases this could screw things up.

src/index.js
```diff
@@ -1,5 +1,8 @@
+import $ from 'jquery';
 import 'bootstrap/dist/js/bootstrap.js';
 
 import 'sass/main.scss';
 
 require('js/App.jsx');
+
+$('[data-toggle="tooltip"]').tooltip();
```
se7entyse7en: mhmm... I'm wondering whether this could cause problems at some point. Related to another comment.

src/js/pages/Home.jsx
dpordomingo: Absolutely. I guess if it locally works bec React just call it if a function is retrieved. Anyhow, calling it explicitly.

src/js/components/pipeline/PullRequests.jsx
dpordomingo: Me too, but we're using `SB Admin 2` which comes with this JQuery thing, so I just let it work.
When facing the PullRequests component, I planned to replace it by any other directly supporting react.

src/index.js
dpordomingo: Same as in the other case. I just used the `SB Admin 2` default one, to be reconsidered later.

dpordomingo:
I think this will be a good candidate for the datatable:
https://github.com/jbetancur/react-data-table-component
which seems to be also configurable.

PR #39, opened 2020-01-28
Change login callback url

se7entyse7en:


se7entyse7en:
@dpordomingo merging since it only changes some urls

PR #42, opened 2020-01-29
Add Development.Breakpoints tool to reveal breakpoints

dpordomingo:
used by #43 and #44

When running the project development (`make serve`), it will appear in the right-bottom corner a small container revealing the current breakpoint based in current bootstrap config.

The compiled version (`make build`) does not add this helper.

### DIFF
--- a/src/js/App.jsx
+++ b/src/js/App.jsx
@@ -6,6 +6,7 @@ import Routes from 'js/Routes';
 import * as serviceWorker from 'js/services/serviceWorker';
 import { Auth0Provider } from 'js/services/react-auth0-spa';
 import history from 'js/services/history';
+import Development from 'js/components/development';
 
 // A function that routes the user to the right place
 // after login
@@ -17,7 +18,7 @@ const onRedirectCallback = appState => {
   );
 };
 
-ReactDOM.render(
+ReactDOM.render(<>
   <Auth0Provider
     domain={window.ENV.auth.domain}
     client_id={window.ENV.auth.clientId}
@@ -25,7 +26,9 @@ ReactDOM.render(
     onRedirectCallback={onRedirectCallback}
   >
     <Routes />
-  </Auth0Provider>,
+  </Auth0Provider>
+  {process.env.NODE_ENV === 'development' && <Development.Breakpoints />}
+</>,
   document.getElementById('root')
 );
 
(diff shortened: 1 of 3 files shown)


src/js/components/development/index.scss
```diff
@@ -0,0 +1,87 @@
...
+        @include media-breakpoint-between(lg,lg) {
+            display: block;
+        }
+    }
+
+    #xl {
+        @include media-breakpoint-up(xl) {
+            display: block;
+        }
+    }
```
se7entyse7en: According to the documentation, it seems your missing the in-between ranges. For example:

- `@include media-breakpoint-between(md, md)` ---> `@media (min-width: 768px) and (max-width: 768px)`
- `@include media-breakpoint-between(lg, lg)` ---> `@media (min-width: 1199.98px) and (max-width: 1199.98px)`

what about values in the middle?

In my understanding you should use these:
```
@include media-breakpoint-only(xs) { ... }
@include media-breakpoint-only(sm) { ... }
@include media-breakpoint-only(md) { ... }
@include media-breakpoint-only(lg) { ... }
@include media-breakpoint-only(xl) { ... }
```

src/js/App.jsx
```diff
@@ -17,15 +18,17 @@ const onRedirectCallback = appState => {
...
   <Auth0Provider
     domain={window.ENV.auth.domain}
     client_id={window.ENV.auth.clientId}
     redirect_uri={window.ENV.auth.redirectUri}
     onRedirectCallback={onRedirectCallback}
   >
     <Routes />
-  </Auth0Provider>,
+  </Auth0Provider>
+  {process.env.NODE_ENV === 'development' && <Development.Breakpoints />}
```
se7entyse7en: This is read from `.env`, right? Is that done automatically? I mean no need for extra deps?

dpordomingo: It's a CreateReactApp built-in env value.
https://create-react-app.dev/docs/adding-custom-environment-variables/
So there is no necessity to do any other thing than `yarn start` as we are already doing.

src/js/components/development/index.scss
dpordomingo: `media-breakpoint-between(a,b)` goes from the lower limit of `a` to the upper limit of `b`, so doing `(a,a)` works from `a` lower limit to its upper one.
But yes, `media-breakpoint-only` is more clear; I don't know why I didn't find it!.
Fixed!

PR #43, opened 2020-01-29
Add login icon and fix Info tooltip

dpordomingo:
- e605325 Fix toltip to be loaded on component mount
- b25a4b8 Add login icon


dpordomingo:
@se7entyse7en you mentioned that you already know a react lib to show tooltips. If you prefer, we can use this PR to add it or postpone it.

se7entyse7en:
@dpordomingo no need now. I'll create a story with two subtasks for removing the usage of jquery.

PR #44, opened 2020-01-29
Integrate the navigation flow 

dpordomingo:
This does not add major visual changes.

The goal of this PR is to have functionality in the stages, so the user will be able to navigate between the Overview and the pipeline stages.

This PR adds these new routes:
- `/stage/:name`, where current `:name` (stages) are: `work-in-progress`, `review`, `merge`, `release`,
- a `Not Found 404` error page.

### Summary of Changes

Easier to be reviewed by commits:
- 32565fa Use `Router.Link` instead of regular `HTML.a` elements. Otherwise, the site would not be a single page application, and there will be full reloads between sections.
- 0c0d986 Make the Pipeline stages attend to URL changes
- fc8b605 Integrate Breadcrumbs to attend to URL changes
- aa871fb Add a Not Found 404 page (using `Simple` template) and make the `Home` to respond only to the exact '/' route which will offer more clarity about what is a not found issue.

Auxiliar change
- 70b9c24 Use a new `Simple` template for `/logout`, `/login` and `/login/callback`. It will provide consistency in the navigation (better UX) and will remove duplicated code (better DX).


se7entyse7en:
Some general thoughts: our public dashboard is supposed to be exposed at `app.athenian.co/demo`, so this implies that we'll have `app.athenian.co/demo/stage/review|merge|...`. But maybe it complicates things if this piece in the middle of the path is "optional" as it will be used only the public dashboard.

So I have multiple options and I don't have a clear picture for some of them so far yet:

- Always use an identifier: `app.athenian.co/some-identifier/stage/...` so on the webapp the path has a "fixed" and well-defined structure. This means that `app.athenian.co` should redirect to `app.athenian.co/some-identifier` according to the logged-in user. So this should be agreed with api (\cc @vmarkovtsev) and with product (\cc @warenlg)
- Don't add any extra path when changing the stage: in this case, we can just verify if the path is `/demo|whatever` and use the gkwillie user to communicate with the api (\cc @vmarkovtsev ). If we want to keep the navigation maybe we could use anchors instead of paths? (\cc @warenlg)
- Maybe it's possible to do some trick at nginx level such as catch `/demo|whatever` route and make the webapp behave as if the `app.athenian.co/demo|whatever` is the root, so the routing in this PR to stages will continue to work, but this may be require some rewriting for fetching static assets at nginx level, but I don't know until I get my hands dirty on this. (\cc @lwsanty)

dpordomingo:
I'd go for a simpler one: deploy de demo under `demo.athenian.app` with its own config.
I don't see the necessity/benefit of relying on `/demo/` url part to change the app behavior.

se7entyse7en:
> I'd go for a simpler one: deploy de demo under `demo.athenian.appz with its own config.

But this has been discussed already during a daily sync about running demo in another cluster (if this is what you mean), and has been discarded due to the need of more ops efforts in terms of maintenance.

se7entyse7en:
@vmarkovtsev could you please remind me what was your plan for making the demo dashboard public? I remember that you mentioned that using an ad-hoc user (such as gkwille) would be easier on the api side instead of permitting to request without an auth header.

But is that fixed user something that is "injected" in the middleware for example by checking the "path"? Or is the webapp that should be aware of this "special user"?

dpordomingo:
For the frontend it should be transparent.

imo clicking in "try as guest", and instead of calling Auth0 for the authentication, call the backend to get an authenticated gets user, and then continue using it regularly.

A simpler alternative could be, on the backend, to use a guest user for no authenticated requests. That way, the frontend will still be able to print data even being not authenticated.

I'd really go for the first because it would let us have different gest users, which could be interesting in the future to let the visitor access the app from different user roles, try different features for them... but if we can not develop this "authenticate as" feature at this moment, we could postpone it, and go with the second one.

se7entyse7en:
> imo clicking in "try as guest", and instead of calling Auth0 for the authentication, call the backend to get an authenticated gets user, and then continue using it regularly.

se7entyse7en:
@warenlg need also you feedback on having a "Login as guest" feature in the login page.

vmarkovtsev:
@se7entyse7en This is how everything works now after the recent PR merges:

- If there is no `Authorization` header, we are @gkwillie. We will assign @gkwillie to be a regular member of the Athenian org so that the public demo is about ourselves.
- If there is an `Authorization` header, we log in users normally.

We can make paths like `app.athenian.co/{account}`. If the user does not belong to the account, API will automatically raise HTTP 401.

se7entyse7en:
Sounds good to me! Thanks! @dpordomingo okay so it's totally ok as you already did to pass empty token 👍

PR #45, opened 2020-01-30
Api client generation

se7entyse7en:
Fixes client generation in Makefile and updates the client.


se7entyse7en:
@dpordomingo I wanted to update the client first before adding bearer auth to make changes clearer.

dpordomingo:
This is autogenerated code, isn't it?

dpordomingo:
if that's the case, LGTM

se7entyse7en:
@dpordomingo yup! autgenerated plus just a fix in path in Makefile after the refactor 👍 
merging then

PR #46, opened 2020-01-30
Add Bearer auth header for api calls

se7entyse7en:
This adds the authentication header with the bearer token required by the api.

For now, it has been added only to `getUser` and `getPipelineDataAPI`. I'd suggest refactoring better the code once we have more calls to the apis. So far I just copy-pasted the same snippet of code for the two above mentioned api calls.

The first commit simply updates the client code according to the updated spec that includes security schema.

### DIFF
--- a/src/js/pages/Home.jsx
+++ b/src/js/pages/Home.jsx
@@ -4,17 +4,33 @@ import Pipeline from 'js/components/pipeline/Pipeline';
 import PullRequests from 'js/components/pipeline/PullRequests';
 import Page from 'js/pages/templates/Page';
 
-import { getPipelineDataInitial, getPipelineDataAPI, getPRs } from 'js/services/api';
+import { getPipelineDataInitial, getPipelineDataAPI, getPRs, fetchApi } from 'js/services/api';
+
+import { useAuth0 } from "../services/react-auth0-spa";
 
 export default () => {
+  const { loading, isAuthenticated, getTokenSilently } = useAuth0();
+
   const [pipelineState, setPipelineData] = useState(getPipelineDataInitial());
   const [prsState, _] = useState(getPRs());
 
   useEffect(() => {
-    getPipelineDataAPI().then(data => {
-      setPipelineData(data);
-    });
-  }, []);
+    const callAPI = async () => {
+      if (loading || !isAuthenticated) {
+        return;
+      };
+
+      getTokenSilently()
+        .then(token => {
+          return fetchApi(token, getPipelineDataAPI);
+        })
+        .then(data => {
+          setPipelineData(data);
+        });
+    };
+
+    callAPI();
+  }, [loading, isAuthenticated, getTokenSilently]);
 
   return (
     <Page>
--- a/src/js/pages/templates/Page.jsx
+++ b/src/js/pages/templates/Page.jsx
@@ -3,18 +3,31 @@ import React, { useState, useEffect } from 'react';
 import Navbar from 'js/components/layout/Navbar';
 import Footer from 'js/components/layout/Footer';
 
-import { getUser } from 'js/services/api';
+import { getUser, fetchApi } from 'js/services/api';
 import { useAuth0 } from 'js/services/react-auth0-spa';
 
 export default ({ children }) => {
+  const { loading, isAuthenticated, getTokenSilently } = useAuth0();
+
   const [userState, setUser] = useState(null);
-  const { isAuthenticated } = useAuth0();
 
   useEffect(() => {
-    if (isAuthenticated) {
-      getUser().then(setUser);
-    }
-  }, []);
+    const callAPI = async () => {
+      if (loading || !isAuthenticated) {
+        return;
+      };
+
+      getTokenSilently()
+        .then(token => {
+          return fetchApi(token, getUser);
+        })
+        .then(data => {
+          setUser();
+        });
+    };
+
+    callAPI();
+  }, [loading, isAuthenticated, getTokenSilently]);;
 
   return (
     <>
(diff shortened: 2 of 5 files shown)


src/js/pages/Home.jsx
```diff
@@ -4,17 +4,33 @@ import Pipeline from 'js/components/pipeline/Pipeline';
...
   const [pipelineState, setPipelineData] = useState(getPipelineDataInitial());
   const [prsState, _] = useState(getPRs());
 
   useEffect(() => {
-    getPipelineDataAPI().then(data => {
-      setPipelineData(data);
-    });
-  }, []);
+    const callAPI = async () => {
+      if (loading || !isAuthenticated) {
```
se7entyse7en: This won't be needed in the future as the page will be closed.

src/js/pages/templates/Page.jsx
```diff
@@ -3,18 +3,31 @@ import React, { useState, useEffect } from 'react';
...
+      if (loading || !isAuthenticated) {
+        return;
+      };
+
+      getTokenSilently()
+        .then(token => {
+          return fetchApi(token, getUser);
+        })
+        .then(data => {
+          setUser();
```
dpordomingo: Trying this locally, it fails because the user is not set.
It should be either:
```javascript
        .then(data => {
￼          setUser(data);
        });
```

or even better:
```javascript
        .then(setUser);
```

src/js/pages/templates/Page.jsx
```diff
@@ -3,18 +3,31 @@ import React, { useState, useEffect } from 'react';
...
   const [userState, setUser] = useState(null);
-  const { isAuthenticated } = useAuth0();
 
   useEffect(() => {
-    if (isAuthenticated) {
-      getUser().then(setUser);
-    }
-  }, []);
+    const callAPI = async () => {
+      if (loading || !isAuthenticated) {
```
dpordomingo: There will be no data if there's no user?
Imo that's not correct.
If there is no user, the token should be just `''` (empty), and let the API answer as `Groundskeeper Willie`

src/js/services/api/openapi-client/ApiClient.js
```diff
@@ -41,6 +41,7 @@ class ApiClient {
          * @type {Array.<String>}
          */
         this.authentications = {
+            'bearerAuth': {type: 'bearer'} // JWT
```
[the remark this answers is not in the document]
se7entyse7en: yup it is 👍

src/js/pages/Home.jsx
```diff
@@ -4,17 +4,33 @@ import Pipeline from 'js/components/pipeline/Pipeline';
...
+
   const [pipelineState, setPipelineData] = useState(getPipelineDataInitial());
   const [prsState, _] = useState(getPRs());
 
   useEffect(() => {
-    getPipelineDataAPI().then(data => {
-      setPipelineData(data);
-    });
-  }, []);
+    const callAPI = async () => {
```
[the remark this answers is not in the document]
se7entyse7en: sorry no need, I was waiting await inside then I changed to promises and forget to remove the async thanks!

src/js/pages/templates/Page.jsx
se7entyse7en: I don't understand, shouldn't the current code produce the same effect of  `.then(setUser);`?

src/js/pages/templates/Page.jsx
se7entyse7en: yup, I put it this way since the way to enable non-auth access is not defined yet, I'd suggest to change it once we decide on that topic. wdyt?

src/js/pages/Home.jsx
```diff
@@ -4,17 +4,33 @@ import Pipeline from 'js/components/pipeline/Pipeline';
...
+      if (loading || !isAuthenticated) {
+        return;
+      };
+
+      getTokenSilently()
```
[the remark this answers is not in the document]
se7entyse7en: depends on the meaning of "ok" 😅 , I mean ofc it cannot show the charts.

src/js/pages/templates/Page.jsx
dpordomingo: I understood (many times) that we'll expose our data to visitors. Won't us?
If so, this PR will solve it:
https://github.com/athenianco/athenian-webapp/pull/47

src/js/pages/templates/Page.jsx
dpordomingo: nope.
Comparing your `line#25` (`setUser()`), with my first alternative (`setUser(data)`), you're passing nothing to `setUser`, so no user is pushed to the component state.

My second alternative does the same as my first one (and cleaner, imho), because instead of passing an anonymous function that calls `setUser(data)` in its body, it passes the `setUser` named function which does the same.

src/js/pages/Home.jsx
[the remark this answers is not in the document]
dpordomingo: ok. Adding a `//TODO` in #47

src/js/pages/templates/Page.jsx
se7entyse7en: > I understood (many times) that we'll expose our data to visitors.

yup, this is correct, but I just don't know yet whether to use the guest user is by providing an empty token or by other means.

dpordomingo: I think that's a good topic for the async discussion in order to keep adding value here (bec imo it's not a big change if we need to take another approach later)

dpordomingo: Also, at this point, it seems that the API is already behaving this same way:
>if there is no auth token, it behaves like using the guest user.

se7entyse7en: yup agree that's why I avoided handling that case, but since it was blocking you did right 👍 thanks!

src/js/pages/templates/Page.jsx
se7entyse7en: ahh ok right sorry! thanks!

PR #52, opened 2020-02-03
Make api url configurable

se7entyse7en:
Signed-off-by: Lou Marvin Caraig <loumarvincaraig@gmail.com>

PR #54, opened 2020-02-03
Add multi-select and date-time filters using API data

dpordomingo:
depends on #52

This PR makes the filters work with data from the API (currently mocked data because the API does not expose this data; see [[ENG-153]])

### To do in the next PR:

Connect the filters to the rest of the App, so it will call the API according to the filter settings.

[ENG-153]: https://athenianco.atlassian.net/browse/ENG-153

### DIFF
--- a/src/js/components/pipeline/TopFilter.jsx
+++ b/src/js/components/pipeline/TopFilter.jsx
@@ -1,46 +1,146 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
+import Select from 'react-select';
+import { timeFormat } from 'd3-time-format';
 
-export default () => (
-    <div className="row mb-3">
-        <div className="col-12">
-            <div className="d-flex justify-content-between align-items-center pb-3 border-bottom">
-                <div className="form-row">
-                    <div className="dropdown col">
-                        <button
-                            className="btn btn-secondary dropdown-toggle bg-white text-xs text-gray-900"
-                            type="button" id="dropdownMenuButton"
-                            data-toggle="dropdown" aria-haspopup="true" aria-expanded="false"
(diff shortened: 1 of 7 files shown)


src/js/components/pipeline/TopFilter.jsx
```diff
@@ -1,46 +1,146 @@
...
+
+import { getRepos, getContributors } from 'js/services/api';
+import { getRepoName, getUserName } from 'js/services/github';
+import AriaLabel from 'js/components/ui/AriaLabel';
+
+export default () => {
+    const [reposState, setRepos] = useState([]);
+    const [areReposReady, setReposReady] = useState(false);
+    const [contributorsState, setContributors] = useState([]);
+    const [areContributorsReady, setContributorsReady] = useState(false);
```
se7entyse7en: Kind of a nit, but isn't it better to uniform the var names such as:
```
cons [whateverFooBar, setWhateverFooBar] = useState(...);
```
or
```
cons [whateverFooBarState, setWhateverFooBar] = useState(...);
```
or
```
cons [whateverFooBarState, setWhateverFooBarState] = useState(...);
```

I think you got the point 😄. I think it makes reading easier, but not a strong opinion.

dpordomingo: I do have a strong opinion about that, and you are right 
I'll uniform this

dpordomingo: I updated all states and state setters from this PR to follow your third option:
```js
const [somethingState, setSomethingState] = useState(something);
```
It's the most verbose, but it's also the more clear one regarding to what is an state, its setter or a regular prop.

PR #50, opened 2020-02-02
Auth handshake webapp <> api

se7entyse7en:
This PR implements the followings:
- add a redirect from url matching the format of the admin invitation link to the login page,
- add the invitation link to the `appState` of Auth0 to make it available in the callback,
- change the Auth0 callback handler to always redirect to `/callback`,
- call `/v1/accept` endpoint of api if the invitation link correspond to an admin link,
- redirect to waiting page on succes.

PR #58, opened 2020-02-06
Improve UX when accepting invitations

dpordomingo:
this will update #50

- Let the user know that is going to accept an invitation,
- Show errors that can happen when accepting an invitation:
  - when trying to accept an invitation being already logged in,
  - when the API is not responding,
  - when the invitation is not valid,
  - when the acceptance process crashed.
- Logout the user if the handshake could not be performed
  (so they can retry)

Other minors:
- Homogenize the new routes with the current ones
- Clean leftover

When trying to accept an invitation being logged:
image

When accepting an invitation, and login is requested:
image

When the invitation is no valid
image

### DIFF
--- a/src/js/Routes.jsx
+++ b/src/js/Routes.jsx
@@ -31,21 +31,20 @@ export default () => {
           <Route path='/stage/:name'>
             <Stage />
           </Route>
-          <Route
-            path='/i/:code(\w{8})'
-            render={props => (
-              <Redirect to={
-                {
-                  pathname: '/login',
-                  state: { inviteLink: window.location.href }
-                }
-              }/>
-            )}
-          />
+          <Route path='/i/:code(\w{8})'>
+            <Redirect to={
+              {
+                pathname: '/login',
+                state: { inviteLink: window.location.href }
+              }
+            } />
+          </Route>
           <Route path='/callback'>
             <Callback />
           </Route>
-          <Route path='/login' component={Login} />
+          <Route path='/login'>
+            <Login />
+          </Route>
           <Route path='/logout'>
             <Logout />
           </Route>
--- a/src/js/pages/auth/Callback.jsx
+++ b/src/js/pages/auth/Callback.jsx
@@ -1,5 +1,4 @@
 import React, { useEffect, useState } from 'react';
-
 import { useLocation, Redirect } from 'react-router-dom';
 
 import { useAuth0 } from 'js/services/react-auth0-spa';
@@ -9,7 +8,7 @@ import { buildApi } from 'js/services/api';
 import InvitationLink from 'js/services/api/openapi-client/model/InvitationLink';
 
 export default () => {
-  const { loading, isAuthenticated, getTokenSilently } = useAuth0();
+  const { loading, isAuthenticated, getTokenSilently, logout } = useAuth0();
   const [redirectTo, setRedirectTo] = useState('');
   const [errorMessage, setErrorMessage] = useState('');
   const query = new URLSearchParams(useLocation().search);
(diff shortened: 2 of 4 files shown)


dpordomingo:
@se7entyse7en This PR is created against your PRed branch `invitation-link-redirect`, so if this is merged, it will update your PR#50 with my proposals.

src/js/Routes.jsx
```diff
@@ -31,21 +31,20 @@ export default () => {
...
+              }
+            } />
+          </Route>
           <Route path='/callback'>
             <Callback />
           </Route>
-          <Route path='/login' component={Login} />
+          <Route path='/login'>
+            <Login />
+          </Route>
```
se7entyse7en: Are you 100% sure that everything still works as expected? Because I seem to recall that I did this because of some issues when reading the query params, but I may not recall correctly.

src/js/pages/auth/Callback.jsx
```diff
@@ -20,31 +19,53 @@ export default () => {
...
+            }
+
+            if (!check.valid || !check.active) {
+              const cause = !check.valid ? 'valid' : 'active';
+              throw new Error(`Invitation is not ${cause}`);
+            }
+
+            let acceptance;
+            try {
+              const acceptance = await api.acceptInvitation(body);
```
se7entyse7en: `acceptance` is not read, also redefined with `const`. I'd just write:
```
try {
    await api.acceptInvitation(body);
}
```

given that we're not reading it.

src/js/pages/auth/Callback.jsx
```diff
@@ -20,31 +19,53 @@ export default () => {
...
+
+          } catch (err) {
+            setErrorMessage(err.message);
+
+            // TODO (dpordomingo):
+            // This is needed because if the invitation fails, the user should not be logged in.
+            // Also, the logout is delayed because Auth0 redirects to the 'logoutRedirectUri'
+            // inmediately, and then the user would not be able to see the error message.
+            // Workaround: store the error, let Auth0 to redirect, and then show the stored message.
+            window.setTimeout(logout, 3000);
```
se7entyse7en: logout({ returnTo: window.ENV.auth.logoutRedirectUri });
```suggestion
            window.setTimeout(function() {
                logout({ returnTo: window.ENV.auth.logoutRedirectUri });
            }, 3000);
```

src/js/pages/auth/Callback.jsx
dpordomingo: I planned to read it, but after testing the functionality I realized that it was not needed and it's now a leftover to delete. Thanks for spotting it!

src/js/Routes.jsx
dpordomingo: I tried the login in all the different ways I could imagine: being logged, not logged and as part of the invitation acceptance workflow. And it worked.

I just followed React Router recommendations about how to define the `<Route>`s, which prefers using children and let the other methods work [for compatibility purposes when hooks are not available](https://reacttraining.com/react-router/web/api/Route/route-render-methods).

Do you remember if you were using hooks (as [now](https://github.com/athenianco/athenian-webapp/pull/58/commits/80fa6a484c387ed9916cbe7af30befed30fca9e0#diff-338bbce5250cb79b92e0ac1bace13e89R10)) when trying this recommended way to define the `<Route>`? Because if you didn't, and tried to pass the `inviteLink` with query params, it could be a reason because it failed.

src/js/pages/auth/Callback.jsx
dpordomingo: why? Since the `Allowed Logout URLs` must be configured at application level, we can rely on that value as the default redirect after the logout (from the [Auth0 logout docs](https://auth0.com/docs/api/authentication?javascript#logout)), can't we?

I also saw it's also suggested that way (without the `returnTo` optional argument) in the [Auth0 React guide](https://auth0.com/docs/quickstart/spa/react#create-the-navbar-component)

se7entyse7en: In theory, there could be more than one `Allowed Logout URLs`, so I think it would be better to specify it explicitly.

src/js/Routes.jsx
se7entyse7en: It could be, probably I tested many ways and in one of them it wasn't working and then I left this as it was before your change. If everything works then the better 👍

src/js/pages/auth/Callback.jsx
dpordomingo: Yes, `Allowed Logout URLs` can contain many different logout urls, and it takes as default the first one according to the docs.

I feel that it would be better to let the default behave like a real default, and explicitly override it when it's really needed.
But if you really prefer to logout with the explicit `redirectTo` just confirm it, and I'll commit.

se7entyse7en: Yup, so that in case we change for example the configuration we don't need to change the webapp.  It would remove the dependency 👍

dpordomingo: That's most probably a good counter-example. If we change the config we'll most probably need to change the web app because the old `redirectTo` would be no longer accepted by the new `Allowed Logout URLs` config.
That would not happen if we would rely on the default behavior.

But since it's an easy change, let's merge this following your preference, and review this in the future if we need it.

dpordomingo:
merged following your suggestions

se7entyse7en:
Thanks a lot!

PR #61, opened 2020-02-14
Invite link

dpordomingo:
closes [[ENG-150]]

### How to test it
In order to create invitations, it is needed to log in from a user being the owner of at least one account. To do so, that user must access the site for the first time through a special (admin) invitation that can be generated from 'athenian-api', running `make invitation-link`

### Internals:
The user was added to the Context, adding its accounts and repos, so from now on it'll be available for the whole app through `js/pages/UserContext::useUser()` ([example](https://github.com/athenianco/athenian-webapp/commit/011f618c24254d0816606e22899b21e6eebbac31#diff-e2170d1e055b911af22853821f913835R9)); it was also done by https://github.com/athenianco/athenian-webapp/pull/60

### Relevant changes (per commit)
- 011f618 Add User to the context
- bf03742 Add [`react-copy-to-clipboard`](https://github.com/nkbt/react-copy-to-clipboard) dependency
- 4d99728 Redesign User menu and generate an invitation from it

### UI

It was followed the latest design.
If the logged user does not own any account, it won't be generated any invite link, nor shown that section in the menu.

image

[ENG-150]: https://athenianco.atlassian.net/browse/ENG-150


src/js/pages/UserContext.jsx
```diff
@@ -0,0 +1,47 @@
+import React, { useState, useEffect, useContext } from 'react';
```
se7entyse7en: Cool!

I'd just move this from `src/js/pages` to `src/js/contexts`. wdyt? I don't know what's the standard for this things.

src/js/pages/UserContext.jsx
```diff
@@ -0,0 +1,47 @@
...
+import { useAuth0 } from 'js/services/react-auth0-spa';
+
+import Simple from 'js/pages/templates/Simple';
+
+import { getUserWithAccountRepos } from 'js/services/api';
+
+const context = React.createContext(null);
```
se7entyse7en: I'd be more explicit as Auth0 is:

```suggestion
const UserContext = React.createContext(null);
```

src/js/pages/UserContext.jsx
```diff
@@ -0,0 +1,47 @@
...
+            setUserState(null);
+            return;
+        };
+
+        getTokenSilently()
+            .then(getUserWithAccountRepos)
+            .then(setUserState)
+            .catch(err => console.error('Could not get user with repos', err));
+
+    }, [loading, isAuthenticated, getTokenSilently]);
```
se7entyse7en: Adding the `getTokenSilently` function shouldn't be necessary.

```suggestion
    }, [loading, isAuthenticated]);
```

src/js/pages/templates/Page.jsx
```diff
@@ -1,32 +1,16 @@
...
-    };
-
-    getTokenSilently()
-      .then(token => fetchApi(token, getUser))
-      .then(setUser)
-      .catch(e => setUser({})); // TODO (dpordomingo):
-    // if could not get the user, but Auth0 is in logged in state, it will set an empty user
-    // We'll handle errors later, but this is blocker to ensure that at least the user can logout.
-  }, [loading, isAuthenticated, getTokenSilently]);;
+  const userContext = useUser();
```
se7entyse7en: Maybe just `user` or `userState`?

```suggestion
  const user = useUser();
```

src/js/services/api/index.js
```diff
@@ -165,6 +176,13 @@ const fetchApiMetricsLine = (api, metrics) => {
   return api.calcMetricsLine(body);
 };
 
+export const getInvitation = async (getTokenFn, accountID) => {
+  const token = await getTokenFn();
+  const api = buildApi(token);
+  const invitation = await api.genInvitation(accountID);
```
se7entyse7en: Let's add error handling.

src/js/services/api/index.js
```diff
@@ -38,18 +38,29 @@ export const getPRs = () => {
   return Array.from(Array(57)).map(() => getRandItem());
 }
 
-export const getUser = (api) => {
-  return api.getUser().then(data => {
-    if (!data.name) {
-      data.name = data.email;
+export const getUserWithAccountRepos = async token => {
+  const api = buildApi(token);
+  const user = await api.getUser();
```
se7entyse7en: Let's add error handling.

src/js/services/api/index.js
```diff
@@ -38,18 +38,29 @@ export const getPRs = () => {
...
+      return { id: Number(accountID), isAdmin, repos: [] };
     }
 
-    return data;
-  }).catch(error => {
-    // TODO(dpordomingo): notify to an error handler which may rise a toast
-    console.error('ERROR calling endpoint', error);
-    throw error;
-  });
+    const reposetsContent = await Promise.all(reposets.map(async reposet => await api.getReposet(reposet.id)));
```
se7entyse7en: Let's add error handling.

src/js/components/layout/Navbar.jsx
```diff
@@ -1,9 +1,13 @@
-import React from 'react';
+import React, { useEffect } from 'react';
+import $ from 'jquery';
```
se7entyse7en: 😢

src/js/components/layout/Navbar.jsx
```diff
@@ -16,26 +20,38 @@ export default ({ user }) => (
...
-  </ul>
-) : (
-    <span className="navbar-nav ml-auto">
-      <Link to="/login">
-        <FontAwesomeIcon icon={faSignInAlt} className="ml-2 text-secondary mr-2 text-gray-400" />
-        Login
+const User = ({ user }) => {
+  useEffect(() => {
+    $(document).on('click', '.nav-item .dropdown-menu', e => e.stopPropagation());
+  }, []);
```
se7entyse7en: Why this? Is the click conflicting with the copy button?

src/js/smart-components/InvitationCreator.jsx
```diff
@@ -0,0 +1,63 @@
...
+        const ownedAccount = getOwnedAccount(user);
+        if (!ownedAccount) {
+            return;
+        };
+
+        setLoadingState(true);
+        getInvitation(getTokenSilently, ownedAccount.id)
+            .then(setUrlState)
+            .finally(() => setLoadingState(false));
+    }, [user, getTokenSilently]);
```
se7entyse7en: Adding `getTokenSilently` shouldn't be needed.

```suggestion
    }, [user]);
```

src/js/smart-components/InvitationCreator.jsx
```diff
@@ -0,0 +1,63 @@
...
+import CopyButton from 'js/components/ui/CopyButton';
+
+import { useUser } from 'js/pages/UserContext';
+
+import { useAuth0 } from 'js/services/react-auth0-spa';
+import { getInvitation } from 'js/services/api';
+
+export default ({ user, className }) => {
+    const { getTokenSilently } = useAuth0();
+    const contextUser = useUser();
```
se7entyse7en: You shouldn't need this as you already have `user`.

src/js/smart-components/InvitationCreator.jsx
```diff
@@ -0,0 +1,63 @@
...
+
+    const [urlState, setUrlState] = useState('');
+    const [loadingState, setLoadingState] = useState(false);
+
+    useEffect(() => {
+        setUrlState('');
```
se7entyse7en: Is this a sort of cleanup for when a user is changed?

src/js/services/api/index.js
```diff
@@ -165,6 +176,13 @@ const fetchApiMetricsLine = (api, metrics) => {
   return api.calcMetricsLine(body);
 };
 
+export const getInvitation = async (getTokenFn, accountID) => {
```
se7entyse7en: Why not accepting the `token` itself as param? It would be more aligned to `getUserWithAccountRepos` as well.

src/js/pages/UserContext.jsx
dpordomingo: Yes, I was also thinking about that. Let's do it!

TL;DR;
When reading about how to structure react apps, I could not find any example in that direction, and I did find some recommendations about not spending too much time thinking about the structure. I realized that this recommendation was based on the fact that as much as the app evolves, it start appearing new kind of elements that might require restructuring the app structure once, and again and again.

src/js/pages/UserContext.jsx
dpordomingo: I have no strong opinion on that.

Auth0 is exporting that context, so it might be interested on qualifying the full name of the exported variable, but we're not. We're exporting [the `useUser()` hook](https://github.com/athenianco/athenian-webapp/pull/61/files/4d99728e505459dfc11007bf9322382189befdb4#diff-1830bf943cca2aa8912122d7f1950409R10) instead.

I would find useful a new naming convention (as we did with `[somethingState, setSomethingState] = userState()`), that could be to suffix that hooks using the context with `Context`; doing this way:
```js
// SomethingContext.jsx
const context = React.createContext();
export const useSomethingContext = () => useContext(context);
```

But if you prefer so, we can also take this other approach:
```js
// SomethingContext.jsx
const SomethingContext = React.createContext();
export const useSomethingContext = () => useContext(context);
```

src/js/pages/UserContext.jsx
dpordomingo: It's required by `react-hooks/exhaustive-deps` lint rule.
(more info at https://github.com/facebook/react/issues/14920#issuecomment-471070149 when answering _"My function value is constant"_)

src/js/pages/templates/Page.jsx
dpordomingo: I initially named it as `user`, but it could be understood as a regular variable, not being used for re-rendering the component when it changes.
`userState` could help to understand that it is not a regular `user` variable, so it will trigger a re-render, but it confuses, in my opinion, the component `state` and the React `context`. That's because I preferred to suffix it with `Context` instead of the regular `State`, used for component state variables.

src/js/services/api/index.js
dpordomingo: It's already done [by its caller](https://github.com/athenianco/athenian-webapp/pull/61/files#diff-1830bf943cca2aa8912122d7f1950409R29).
About making the error handling smarter (regarding UX and DX), I'd do it as part of _"Error handling with toasts"_ from our "webapp nice to have checklist"

src/js/services/api/index.js
dpordomingo: _ditto_

src/js/components/layout/Navbar.jsx
dpordomingo: Yes :cry:. As soon as we have a replacement for the dropdown menu, we'll get rid of this.
By the way. You mentioned you already know for a good candidate for replacing the current jquery based tooltips with react ones (but you didn't mention which one). Do you already know about a react lib for replacing current jquery dropdowns?

src/js/smart-components/InvitationCreator.jsx
dpordomingo: It seems that the token is fetched by Auth0 `getTokenSilently`, who takes care of renewing it if needed. I understood that is the reason because we're not storing the user token into the user itself, and we're relying in Auth0 `getTokenSilently` instead.

I'd propose some improvements on this, in order to inject that service into the api so it would be transparent for the `useEffect`. But it's something that I'd not handle at this stage, and I'd postpone once we release v1 instead.

src/js/smart-components/InvitationCreator.jsx
dpordomingo: yes.

src/js/smart-components/InvitationCreator.jsx
dpordomingo: _ditto_

src/js/services/api/index.js
dpordomingo: done by 9469b7d

src/js/components/layout/Navbar.jsx
dpordomingo: SB Admin 2 dropdown is using Bootstrap dropdown, which is being configured to be closed when you click in the dropdown (see [sources](https://github.com/twbs/bootstrap/blob/master/js/src/dropdown.js#L523))
Then, no matter if you click in the user picture, in the logout link, or in the copy button, the dropdown is automatically closed as shown in their sources.

Reviewing the sources, I realized that the dropdown is not automatically closed when clicking INSIDE a `<form>` contained in the dropdown (which makes sense, because if it would be closed when you click there, you would not be able to fill the form inside a dropdown).

Then I would purpose this "trick" fc4b61b (without using jQuery), in order to avoid closing the dropdown using that undocumented feature.

src/js/components/layout/Navbar.jsx
dpordomingo: aside.
I see two kinds of usage of `jquery`.
- Using it to render things... which could conflict with react renders if not used properly.
- Using it to select elements or adding events not tied to react behavior.
In this case, we're using it for the second use-case (anyhow, see below for [a proposed workaround](https://github.com/athenianco/athenian-webapp/pull/61#discussion_r380078323)).

src/js/services/api/index.js
dpordomingo: Done by its [caller](https://github.com/athenianco/athenian-webapp/pull/61/files#diff-4dc410104b4f68088afe318ed78d30b8R30).

dpordomingo:
Comments addressed.
Thanks for the review!

src/js/pages/UserContext.jsx
dpordomingo: dd73c06 moved `Auth0` context to the same place.

se7entyse7en: The recommendation you found makes sense, especially for an early stage webapp. But since putting all the contexts in one place is not that much effort I think that helps having things clearer. 👍

Thanks for this!

src/js/pages/UserContext.jsx
se7entyse7en: I don't have a strong opinion either. Up to you 👍, let's just be consistent once we choose one (and I know I'll be the one that will forget what's the rule... 😅). Probably I know what you're thinking, at some point, we need to write down these internal conventions.

src/js/pages/UserContext.jsx
se7entyse7en: Interesting, thanks for pointing to the very specific section of that very long thread.

src/js/pages/templates/Page.jsx
se7entyse7en: > userState could help to understand that it is not a regular user variable, so it will trigger a re-render, but it confuses, in my opinion, the component state and the React context

I agree on this.

> That's because I preferred to suffix it with Context instead of the regular State, used for component state variables.

By thinking more about this it totally make sense. I don't know the best practice, but I'm wondering whether it makes even more sense to make the value of the context not the user itself, but an object with the user. So that we'll actually end up doing:

```
const { user } = userUser();
```

src/js/services/api/index.js
se7entyse7en: Agree 👍 thanks!

src/js/components/layout/Navbar.jsx
se7entyse7en: > You mentioned you already know for a good candidate for replacing the current jquery based tooltips with react ones (but you didn't mention which one).

Right, 😅. I didn't do an extensive research though, but this seems pretty flexible and well-maintained: https://github.com/wwayne/react-tooltip.

Regarding dropdowns, nope.

Let's add the this lib in the checklist and the dropdowns as well. 👍

src/js/components/layout/Navbar.jsx
se7entyse7en: LGTM 👍, nice find!

src/js/components/layout/Navbar.jsx
se7entyse7en: Added.

src/js/smart-components/InvitationCreator.jsx
se7entyse7en: > I'd propose some improvements on this, in order to inject that service into the api so it would be transparent for the useEffect.

I like this, it totally make sense to make it transparent.

> But it's something that I'd not handle at this stage, and I'd postpone once we release v1 instead.

Makes sense 👍.

But this comment was about the `contextUser`. I mean in line 20, you doing:
```
if (!contextUser)
```
but isn't it the same doing:
```
if (!user)
```
?

se7entyse7en:
Done second pass, just a couple of minor things 👍

src/js/pages/UserContext.jsx
dpordomingo: Applying the second a051bb5

```js
// js/context/Something.jsx
const SomethingContext = React.createContext();
export const useSomethingContext = () => useContext(SomethingContext);
```

src/js/pages/templates/Page.jsx
dpordomingo: As far as reading the variable you know that it comes from the outside, and it will be updated when it changes (because of a re-render), it lgtm both:
```jsx
const pageSize = 2;
const userContext = useUserContext(); 
// const {userContext} = useUserContext(); // whatever you prefer, idm.
const {prsState, setPRsState} = useState([]);
// ...
// many lines below...
<span>{userContext.name}</span>
{prsState.splice(0,pageSize).map(pr=>pr.title)}

```

src/js/components/layout/Navbar.jsx
dpordomingo: looks good!

src/js/smart-components/InvitationCreator.jsx
dpordomingo: Ouch! I don't know what happened, that I answered something absolutely unrelated to your comment.
Yes, you're right; I already delete it https://github.com/athenianco/athenian-webapp/pull/61/commits/ed7ace49596536b8c8b153b733203f3c4f67d047 because I also discovered it.

dpordomingo:
- https://github.com/athenianco/athenian-webapp/pull/61#discussion_r380598598 Applying new Context convention
- https://github.com/athenianco/athenian-webapp/pull/61#discussion_r380604260 I have not a preference on how to return the context content as far it's suffixed with "Context" (or any other similar to understand where is it coming from)
- https://github.com/athenianco/athenian-webapp/pull/61#discussion_r380606739 `userContext` was already removed since `user` is already passed as a component argument.

dpordomingo:
After reviewing pending things I see that all the pending suggestions were already addressed, so rebasing, squashing and merging this.
If you want to choose any other naming convention regarding to React Context, and how to serve its content (https://github.com/athenianco/athenian-webapp/pull/61#discussion_r380604260), feel free to do so, and I'll PR it in a separate PR.

PR #62, opened 2020-02-15
add required static field to be sent over pubsub

lwsanty:
blocked by https://github.com/athenianco/ci/pull/4

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ stages:
   - name: test
   - name: build
   - name: deploy
-    if: branch = master AND type != pull_request
+    if: branch IN (master, production) AND type != pull_request
 
 jobs:
   fast_finish: true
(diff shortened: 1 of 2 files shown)


lwsanty:
@dpordomingo sure
having a new CD requires us having different [pubsub message](https://github.com/athenianco/ci/pull/4/files) where `SERVICE_NAME` is the name of the service that's deployment is required to be updated. Can be either `api` or `web-app`.

se7entyse7en:
> blocked by athenianco/ci#4

Actually not blocked, right? If we merge this PR it will be just unused. And we need to merge CI once we already have the CD up and running otherwise we won't be able to deploy. Or did I miss something?

lwsanty:
@se7entyse7en yes we won't be able to deploy and Travis job will fail because service was not set

PR #64, opened 2020-02-20
Enhance navigation

dpordomingo:
blocks https://github.com/athenianco/athenian-webapp/pull/60

## Changes:

- eaeb80e Convert `/stage/:name` into a nested route of `/stage`
    It will let Stage to preserve some parts untouched (not re-rendered) when navigating between overview and stages

- 8dec13c Add `useBreadcrumbsContext` hook to change the Breadcrumbs from Page parts
    Page is wrapped inside a `BreadcrumbsContext` which stores a setter for
    the Breadcrumbs to be used by Page parts through `useBreadcrumbsContext`.

### DIFF
--- a/src/js/Routes.jsx
+++ b/src/js/Routes.jsx
@@ -13,7 +13,9 @@ import PrivateRoute from 'js/PrivateRoute';
 
 import Home from 'js/pages/Home';
 import UserContext from 'js/context/User';
+import Pipeline from 'js/pages/Pipeline';
 import Stage from 'js/pages/pipeline/Stage';
+import Overview from 'js/pages/pipeline/Overview';
 import Login from 'js/pages/auth/Login';
 import Logout from 'js/pages/auth/Logout';
 import Callback from 'js/pages/auth/Callback';
@@ -30,8 +32,20 @@ export default () => {
             <Route exact path='/'>
               <Home />
             </Route>
-            <Route path='/stage/:name'>
-              <Stage />
+            <Route path='/stage/:name?'>
+              <Pipeline>
+                <Switch>
+                  <Route exact path='/stage'>
+                    <Redirect to='/stage/overview' />
+                  </Route>
+                  <Route exact path='/stage/overview'>
+                    <Overview />
+                  </Route>
+                  <Route path='/stage/:name'>
+                    <Stage />
+                  </Route>
+                </Switch>
+              </Pipeline>
             </Route>
             <Route path='/i/:code(\w{8})'>
               <Redirect to={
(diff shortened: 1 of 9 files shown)


src/js/pages/Home.jsx
```diff
@@ -1,5 +1,4 @@
 import React from 'react';
+import { Redirect } from 'react-router-dom';
 
-import Overview from 'js/pages/pipeline/Overview';
-
-export default () => <Overview />;
+export default () => <Redirect to="/stage" />
```
se7entyse7en: Depending on the route comment, this would need to change accordingly.

src/js/pages/Pipeline.jsx
```diff
@@ -0,0 +1,44 @@
+import React, { useState, useEffect } from 'react';
+import { useParams } from "react-router-dom";
```
se7entyse7en: Quotes

src/js/pages/templates/Page.jsx
```diff
@@ -1,25 +1,26 @@
...
-import { useUserContext } from 'js/context/User';
 
-export default ({ breadcrumbs, children }) => {
+export default ({ children }) => {
   const userContext = useUserContext();
+  const [breadcrumbsState, setBreadcrumbsState] = useState(null);
 
   return (
-    <>
+    <BreadcrumbsContext setBreadcrumbsFn={setBreadcrumbsState}>
```
se7entyse7en: I still don't know whether using a context for this case is an overkill, but I don't have a strong opinion. I mean it's used only for making a setter global to the children.

dpordomingo: It's a way to control the thumbnails from everywhere. As soon as we start having more paths, it will free us from passing the Breadcrumbs down everywhere

dpordomingo:
`/stage` now redirects to `/stage/overview` ([see](https://github.com/athenianco/athenian-webapp/pull/64/files#diff-be04a9bedbc0e0c929fb112e61ff0c11R39))
Is this ready to be merged?

src/js/pages/Home.jsx
dpordomingo: done

src/js/Routes.jsx
```diff
@@ -30,8 +32,17 @@ export default () => {
             <Route exact path='/'>
               <Home />
             </Route>
-            <Route path='/stage/:name'>
-              <Stage />
+            <Route path='/stage/:name?'>
+              <Pipeline>
+                <Switch>
+                  <Route exact path='/stage'>
```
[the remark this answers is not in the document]
dpordomingo: done

PR #66, opened 2020-02-24
Integrate Main Metrics

dpordomingo:
depends on https://github.com/athenianco/athenian-webapp/pull/60
depends on https://github.com/athenianco/athenian-webapp/pull/65

#relevant changes
- cda64c5 Add PRs to the context
  PRs will be needed in many parts of the app, so moving it to the context.
- afc75f2 Use PRs from context in PRs table
- 9fa8826 Make MainMetrics component configurable, and use some real data
- cc79341 Add LeadTime to the fetched metrics and use it in MainMetrics
 Since the LeadTime won't be served separately for the metrics, but from the /metrics_line endpoint,
it will be fetched with the rest of the metrics for the pipeline. Doing so, forces us to re organize
the data because rendering was pretty coupled to the structure of the data served by the API.

### autogenerate code
- bbc20cc Update autogenerated OpenAPI client from Athenian API scheme


src/js/pages/pipeline/Body.jsx
```diff
@@ -0,0 +1,105 @@
...
+    const { getTokenSilently } = useAuth0();
+    const userContext = useUserContext();
+
+    const [pipelineState, setPipelineState] = useState([]);
+    const [leadTimeState, setLeadTimeState] = useState({});
+    const { name } = useParams()
+    const activeStageState = pipelineStagesConf.findIndex(metric => metric.slug === name);
+
+    const { dateInterval, repositories, contributors } = useFiltersContext();
+    const { prs } = usePRsContext();
```
se7entyse7en: IMO we're starting abusing the usage of react contexts. I don't see the advantage in this case of using react context instead of simply passing the data as a prop. The main reason for using contexts is to avoid having props to drill down a whole chain of components, but this is not the case. We shouldn't abuse them, as theoretically everything could be treated as part of a context and we avoid passing props at all.

src/js/pages/pipeline/Body.jsx
```diff
@@ -0,0 +1,105 @@
...
+        <>
+            {withMainMetrics(leadTimeState, prs, dateInterval)}
+            <Thumbnails stages={pipelineState} activeCard={activeStageState} />
+            {children}
+        </>
+    );
+};
+
+const withMainMetrics = (leadTimeData, prs, dateInterval) => {
+    const prCount = prs.filter(pr => (pr.created >= dateInterval.from && pr.created <= dateInterval.to)).length;
```
se7entyse7en: I don't get this, the PRs returned by the api are already filtered by the interval requested, aren't they? \cc @vmarkovtsev

vmarkovtsev: Yes, they are filtered. What's more, the PR creation time does not have to fit into the time interval. In fact, it is the PR closure time that must fit instead. I wonder about this condition, too.

dpordomingo: From the [product docs](https://docs.google.com/document/d/19uHBCKmqwYcysiIZodR_y37qxpG9nyP9GPpZmCGV3Pg)
image
>_Number of Pull Requests that have been **created** according to the values of the 3 filters._

As Vadim said, the PRs returned by the API, are not necessarily created in that date interval.

src/js/pages/pipeline/Body.jsx
dpordomingo: Thanks to `{children}`, there are components which do not know what other component is being rendered inside, so that components can not pass `prs` down.
Initially, I worked with the idea of having all the data for each component calculated by each different API endpoint, but turns out that the frontend needs to build as much as possible on their own; then, since the PRs are needed almost everywhere, and some components do not know what's inside, PRs became a sort of thing potentially needed everywhere.

Anyhow, I also feel that PRs are maybe too much to be in the Context, but if you don't really oppose we can consider this as a part of the technical debt to be reviewed later when we could also have a better understanding on how everything is working.

src/js/pages/pipeline/Body.jsx
vmarkovtsev: I am a bit out of context here. Is this about listing WIP PRs? Then the returned list should be fine as-is. Or are you requesting all the stages at once? Then how about filtering by the WIP stage?

vmarkovtsev: Ah, right, the PR can be created and already transition to another stage. Yep, this is correct.

dpordomingo: image
Considering that Main Metric, the definition of the "Pull Request in progress" is "the PRs that was CREATED in that period".
The PRs returned by `/filter/pull_requests` for certain timeFrame are those that were MODIFIED in that date frame, so they can be created before that timeFrame.

vmarkovtsev: @warenlg We should perhaps reconsider this definition. The majority of PRs will be created before the selected time period.

src/js/pages/pipeline/Body.jsx
se7entyse7en: I agree in post-poning at least post mvp. I got your point, but probably we could find an easier alternative since having many nested context it's easy from the development point of view as you have everything you need, but it becomes harder from the debugging point of view. Let's keep this tech-debt in mind.

dpordomingo: Yes, I agree that it can complicate some things later and that we can find a better solution that would solve both necessities.
I'll revisit this after MVP release and once I have a deeper knowledge of react arch.

dpordomingo:
Merging this since all mayor concerns were already answered.

*TODO:*
- To review the usage of the React Context, especially the necessity of storing the PRs there.

src/js/pages/pipeline/Body.jsx
warenlg: Sorry guys, my bad on this one. I forgot to update this line when iterating with Zuri on the design. I just clarified the doc. But I put it here again:
```
They are 3 main sticky metrics. They should give a first impression of the global state of the software delivery pipeline:
Lead Time: How fast are we delivering software? E.g. 11 days
Time elapsed between the creation of the 1st commit in a Pull Request and the code being used in production.
~~Pull Requests in Progress~~ Volume: What volume of work the team has been dealing with? E.g. 46 pull requests
Number of Pull Requests that have at least one living stage according to the values of the 3 filters.
Engagement: How engaged is the team in this sprint? E.g 9 contributors.
Number of people engaged according to the values of the 3 filters. (could be a committer, an author of one comment, anything)
```

se7entyse7en: okay, so this means no need for an extra filtering on webapp side \cc @dpordomingo

dpordomingo: The brand new logic will be delivered by https://github.com/athenianco/athenian-webapp/pull/76

PR #69, opened 2020-02-24
Calculate variation between current and prev period

dpordomingo:
It will be called the api doubling the passed DateInterval.
Once the response is obtained, it will be calculated the variation between the previous period, and the passed DateInterval.
Once obtained the variation, it's returned the data only for the passed DateInterval.

### DIFF
--- a/src/js/components/ui/Badge.jsx
+++ b/src/js/components/ui/Badge.jsx
@@ -7,6 +7,7 @@ import { faAngleDown, faAngleUp, faEquals } from '@fortawesome/free-solid-svg-ic
 export default ({ value, trend, className }) => {
   const commonClasses = ['badge', 'font-weight-normal', 'align-middle', 'd-inline-block'];
   let customClasses, icon;
+  let suffix = '%';
   if (!trend) {
     customClasses = ['badge-pill', 'badge-secondary', 'py-1', 'px-2'];
   } else if (value < 0) {
@@ -18,14 +19,15 @@ export default ({ value, trend, className }) => {
   } else {
     customClasses = ['badge-secondary'];
     icon = faEquals;
+    suffix = '';
   }
 
   return (
     <span className={classnames('badge', className, ...commonClasses, ...customClasses)}>
       {trend ? (
         <>
-          <FontAwesomeIcon icon={icon} className="align-bottom mr-1" />
-          {value < 0 ? -value : value}%
+          <FontAwesomeIcon icon={icon} className="align-bottom" />
+          {value ? <span className="ml-1">{Math.abs(value)}{suffix}</span> : ''}
         </>
       ) : (
           <>{value}</>
--- a/src/js/services/api/index.js
+++ b/src/js/services/api/index.js
@@ -100,29 +100,68 @@ export const getContributors = (token, userAccount, from, to, repos) => {
 
 export const getMetrics = (api, accountId, dateInterval, repos, contributors) => {
   const metrics = ['lead-time', 'wip-time', 'review-time', 'merging-time', 'release-time'];
-  const data = metrics.reduce((acc, metric) => {
-    acc[metric] = { data: [], avg: 0, sum: 0 };
-    return acc;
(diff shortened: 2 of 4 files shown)


src/js/services/api/index.js
```diff
@@ -67,83 +98,82 @@ export const getContributors = (token, userAccount, from, to, repos) => {
...
-          if (!thumbsData[metricNo]) {
-            thumbsData[metricNo] = [];
-          }
-
-          thumbsData[metricNo].push({ x: new Date(day.date), y: dateTime.secondsToHours(value) });
-        });
-      });
+export const getMetrics = (api, accountId, dateInterval, repos, contributors) => {
+  const metrics = ['lead-time', 'wip-time', 'review-time', 'merging-time', 'release-time'];
+  const result = metrics.reduce((acc, metric) => {
```
se7entyse7en: No need to reduce here, you're not aggregating you're just doing initialization.

src/js/services/api/index.js
```diff
@@ -67,83 +98,82 @@ export const getContributors = (token, userAccount, from, to, repos) => {
...
+  };
 
+  return fetchApiMetricsLine(api, metrics, accountId, doubleDateInterval, repos, contributors).then(apiData => {
+    if (!apiData.calculated || !apiData.calculated[0]) {
+      throw new Error('API returned no calculated data');
     }
 
-    return getPipelineThumbs(thumbsData);
+    // This will unfold values, to have separate data per different metric, instead of all metrics under the same day.
+    apiData.calculated[0].values.forEach(step => {
```
se7entyse7en: Could you remind me when this happens `len(calculated) > 1`?

src/js/services/api/index.js
```diff
@@ -67,83 +98,82 @@ export const getContributors = (token, userAccount, from, to, repos) => {
...
+      step.values.forEach((value, metricNo) => {
+        if (new Date(step.date) >= dateInterval.from) {
```
se7entyse7en: The date can be initialized once outside the `forEach`.

se7entyse7en:
@dpordomingo I reviewed only the commit that is part of this PR as you said in the PR desc.

src/js/components/ui/Badge.jsx
```diff
@@ -7,25 +7,29 @@ import { faAngleDown, faAngleUp, faEquals } from '@fortawesome/free-solid-svg-ic
 export default ({ value, trend, className }) => {
   const commonClasses = ['badge', 'font-weight-normal', 'align-middle', 'd-inline-block'];
   let customClasses, icon;
+  let suffix = '%';
   if (!trend) {
     customClasses = ['badge-pill', 'badge-secondary', 'py-1', 'px-2'];
   } else if (value < 0) {
     customClasses = ['badge-danger'];
+    value = -value;
```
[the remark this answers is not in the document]
dpordomingo: more clear at least!.

src/js/services/api/index.js
dpordomingo: I must recognize that I use `reduce` too much, and you found a way to reduce to absurdity the way I use it that much.

src/js/services/api/index.js
dpordomingo: The API returns as much `calculated` items, as `for` filters you pass the `/metrics_line` endpoint.
The webapp is currently using only one as you can see in [`api/fetchApiMetricsLine` method sources](https://github.com/athenianco/athenian-webapp/blob/master/src/js/services/api/index.js#L140)

Once we refactor the API as we have in our wishlist, that method will be also affected, in order to reinforce that logic.

src/js/services/api/index.js
dpordomingo: optimizing cpu cycles! that's neat! thanks!

dpordomingo:
Thanks for revieweing @se7entyse7en !
I addressed your suggestions by 7eca21696766322fbb0b44392fc66977ed629562

PR #70, opened 2020-02-24
Show PRs in each pipeline thumbnail

dpordomingo:
depends on #60
depends on #65
depends on #66
depends on #67
depends on #68
depends on #69

changes are only introduced by 17ad96c


dpordomingo:
@warenlg This PR will reflects the original definition for this thumbnail badge.

There is still an [ongoing conversation about it in our Prod docs](https://docs.google.com/document/d/19uHBCKmqwYcysiIZodR_y37qxpG9nyP9GPpZmCGV3Pg/edit?disco=AAAAJIIvMxc), explaining that it should measure the number of PRs that transitioned each stage in the period of time being analyzed. If that other definition is approved, it should be considered separately.

Issue #72, opened 2020-02-24
Adding the proper typography and type styles override


znegrin:
I will add the Cerebri typography to the webapp and override the styles to set up the proper sizes [according to the prototypes](https://zurinegrn558160.invisionapp.com/overview/Athenian---Dashboard-ck6gn9es0003g01ab2k2y2k3p/screens).

znegrin:
@dpordomingo I have several doubts about how to proceed with this in terms of how to override the existing work. Please feel free to ping me here or at Slack so I can start working on it properly.

Thank you!

dpordomingo:
About typography, for example, SB Admin 2 defines its fonts [with `$font-family-sans-serif` variable](https://github.com/BlackrockDigital/startbootstrap-sb-admin-2/blob/master/scss/_variables.scss#L40), which could be overrided from `src/sass/_overrides.scss` as done [in this other situations](https://github.com/athenianco/athenian-webapp/blob/master/src/sass/_overrides.scss#L29).

znegrin:
These changes are made in #74.

PR #74, opened 2020-02-24
Basic typography additions and changes to match the prototypes

znegrin:
Based on the issue #72 I added the Cerebri web font and made minor size changes. In a future issue I will continue improving further typographic details.


src/sass/_typography.scss
```diff
@@ -1 +1,13 @@
-@import url('https://fonts.googleapis.com/css?family=Nunito:200,200i,300,300i,400,400i,600,600i,700,700i,800,800i,900,900i&display=swap');
+// @import url('https://fonts.googleapis.com/css?family=Nunito:200,200i,300,300i,400,400i,600,600i,700,700i,800,800i,900,900i&display=swap');
```
dpordomingo: If we're not longer using it, we can delete it ;)

yarn.lock
```diff
@@ -3596,7 +3596,7 @@ debug@2.6.9, debug@^2.2.0, debug@^2.3.3, debug@^2.6.0, debug@^2.6.9:
   dependencies:
     ms "2.0.0"
 
-debug@^3.0.0, debug@^3.1.0, debug@^3.1.1, debug@^3.2.5, debug@^3.2.6:
+debug@^3.0.0, debug@^3.1.0, debug@^3.1.1, debug@^3.2.5:
```
dpordomingo: Can you check out (not commit)this file? I know that it's changed after running `make install`, which is needed to start the app, but we should not commit it this way. (I'll try to fix this inconvenience once I have time)

znegrin: I tried to revert this but I think I didn't do it the right way as the file keeps there, but now with a conflict. Not sure about how to check it out.

PR #75, opened 2020-02-25
Fetch the API spec from the Git repo

vmarkovtsev:


Makefile
```diff
@@ -16,19 +16,26 @@ $(MAKEFILE):
...
 
 API_SERVICE_PATH := src/js/services/api
 API_CLIENT_CODE_PATH := $(API_SERVICE_PATH)/openapi-client
+API_SPEC_PATH := $(API_SERVICE_PATH)/openapi.$(API_VERSION).yaml
+API_CLONE_PATH ?= /tmp/athenian-api
 
 OPENAPI_GENERATOR := node_modules/@openapitools/openapi-generator-cli/bin/openapi-generator
 
+$(API_SPEC_PATH):
+	rm -rf $(API_CLONE_PATH)
```
vmarkovtsev: We need rimrafs both at the beginning and at the end because if `git checkout` fails, the repo will remain on disk and the repeated call to `make api-generate` will fail.

src/js/services/api/openapi-client/ApiClient.js
```diff
@@ -32,9 +32,9 @@ class ApiClient {
         /**
          * The base URL against which to resolve every API call's (relative) path.
          * @type {String}
-         * @default https://api.athenian.co/v1
+         * @default http://}/v1
          */
-        this.basePath = 'https://api.athenian.co/v1'.replace(/\/+$/, '');
+        this.basePath = 'http://}/v1'.replace(/\/+$/, '');
```
vmarkovtsev: This is not looking good but I don't think we are using this value really.

Makefile
```diff
@@ -16,19 +16,26 @@ $(MAKEFILE):
 # CI is set to '1' in travis by default
 CI ?=
 
-# TODO (dpordomingo): consider locking to fixed version once we start doing releases.
-API_SPECS_URL ?= https://raw.githubusercontent.com/athenianco/athenian-api/master/server/athenian/api/openapi/openapi.yaml
+API_VERSION ?= v0.0.39
 
 API_SERVICE_PATH := src/js/services/api
 API_CLIENT_CODE_PATH := $(API_SERVICE_PATH)/openapi-client
+API_SPEC_PATH := $(API_SERVICE_PATH)/openapi.$(API_VERSION).yaml
```
dpordomingo: Make this default, so the user will be able to run:
```
API_SPEC_PATH=/my/development/openapi.yml make generate
```

.gitignore
```diff
@@ -1,3 +1,5 @@
+src/js/services/api/openapi.*.yaml
```
dpordomingo: I think it could be very useful if we commit this file.
wdyt?

vmarkovtsev: You will soon have hundreds of such files - they are generated for each tag.

src/js/services/api/openapi-client/ApiClient.js
vmarkovtsev: Yes, it is autogenerated.

yarn.lock
```diff
@@ -3596,7 +3596,7 @@ debug@2.6.9, debug@^2.2.0, debug@^2.3.3, debug@^2.6.0, debug@^2.6.9:
   dependencies:
     ms "2.0.0"
 
-debug@^3.0.0, debug@^3.1.0, debug@^3.1.1, debug@^3.2.5, debug@^3.2.6:
+debug@^3.0.0, debug@^3.1.0, debug@^3.1.1, debug@^3.2.5:
```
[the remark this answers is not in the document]
vmarkovtsev: ok

Makefile
```diff
@@ -16,19 +16,26 @@ $(MAKEFILE):
...
+API_CLONE_PATH ?= /tmp/athenian-api
 
 OPENAPI_GENERATOR := node_modules/@openapitools/openapi-generator-cli/bin/openapi-generator
 
+$(API_SPEC_PATH):
+	rm -rf $(API_CLONE_PATH)
+	git clone --single-branch --no-checkout https://github.com/athenianco/athenian-api $(API_CLONE_PATH)
```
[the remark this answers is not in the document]
vmarkovtsev: `--depth=1` is going to drop all the tags except the latest.

vmarkovtsev:
@dpordomingo Done

Makefile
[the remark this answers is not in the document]
dpordomingo: you can disregard my comment. I thought I had delete it. It seems I didn't.

.gitignore
dpordomingo: what about deleting the old one once the new api is autogenerated?

dpordomingo: Ok, this is not a blocker, just a nice to have, so you can PR it separately if you feel happy ;)

vmarkovtsev: Well... This is possible I guess, but we will need to compare version strings in Shell.

Makefile
dpordomingo: or...
```
old_spec=`ls -1 src/js/services/api/openapi.*.yaml`
# clone new
# move new into $(API_SERVICE_PATH)
# generate the new api
rm $(old_spec)
```

.gitignore
dpordomingo: or...
```
old_spec=`ls -1 src/js/services/api/openapi.*.yaml`
# clone new
# move new into $(API_SERVICE_PATH)
# generate the new api
rm $(old_spec)
```

PR #77, opened 2020-02-25
/bearer

vmarkovtsev:
This adds the following:

image

I need this page to quickly debug the API calls that require authorization.

Depends on #75

### DIFF
--- a/src/js/Routes.jsx
+++ b/src/js/Routes.jsx
@@ -18,6 +18,7 @@ import Stage from 'js/pages/pipeline/Stage';
 import Overview from 'js/pages/pipeline/Overview';
 import Login from 'js/pages/auth/Login';
 import Logout from 'js/pages/auth/Logout';
+import Bearer from 'js/pages/auth/Bearer';
 import Callback from 'js/pages/auth/Callback';
 import Waiting from 'js/pages/Waiting';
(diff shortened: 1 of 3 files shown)


dpordomingo:
Why not using the chrome inspector?

image

The bearer is also there, and it's easy to copy and paste. And you can even copy the whole curl request from there!

vmarkovtsev:
This was the intention. Using devtools is too complex. I know how to use it, but tokens get expired quick and going to devtools, refreshing the page, locating the API call, viewing the headers is too slow.

dpordomingo:
I don't know how often these tokens expire, but I had a devtool already opened from almost an hour ago, and I replayed one old random XHR request, and it succeeded, so it seems that it has a long life span.

But anyhow, if you really, really, REALLY need this, then let's add it as a development tool [as done with `Development.Breakpoints`](https://github.com/athenianco/athenian-webapp/blob/master/src/js/App.jsx#L28)
- move the new component to be a sibling of the already existent `Development.Breakpoints`, and
- move the css there.

vmarkovtsev:
You are challenging my ReactJS skills! Challenge accepted :sunglasses:

vmarkovtsev:
Slack discussion: https://athenian.slack.com/archives/CQUFEUP5H/p1582643124047900

PR #78, opened 2020-02-25
Add /bearer route to generate auth tokens

vmarkovtsev:
Close #77
Alternative to #77. Follows https://github.com/athenianco/athenian-webapp/pull/77#issuecomment-590908969

However, it does not work and I cannot find the reason why.

image

### DIFF
--- a/src/js/Routes.jsx
+++ b/src/js/Routes.jsx
@@ -18,12 +18,15 @@ import Stage from 'js/pages/pipeline/Stage';
 import Overview from 'js/pages/pipeline/Overview';
 import Login from 'js/pages/auth/Login';
 import Logout from 'js/pages/auth/Logout';
+import Development from 'js/components/development';
 import Callback from 'js/pages/auth/Callback';
 import Waiting from 'js/pages/Waiting';
 import NotFound404 from 'js/pages/NotFound404';
 import Prototypes from 'js/pages/prototypes/Prototypes';
 
 export default () => {
+
+  const devmode = process.env.NODE_ENV === 'development';
 
   return (
     <React.StrictMode>
@@ -65,6 +68,7 @@ export default () => {
             <Route path='/logout'>
               <Logout />
             </Route>
+            {devmode && <Route path='/bearer'><Development.Bearer /></Route>}
             <PrivateRoute path='/waiting' component={Waiting} />
             <Route path='/prototypes'>
               <Prototypes />
--- a/src/js/components/development/index.jsx
+++ b/src/js/components/development/index.jsx
@@ -1,4 +1,6 @@
-import React from 'react';
+import React, {useState} from 'react';
+import { useAuth0 } from 'js/context/Auth0';
+import Simple from 'js/pages/templates/Simple';
 import 'js/components/development/index.scss';
 
 export default {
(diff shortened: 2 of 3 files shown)


src/js/Routes.jsx
```diff
@@ -65,6 +68,7 @@ export default () => {
             <Route path='/logout'>
               <Logout />
             </Route>
+            {devmode && <Route path='/bearer'><Bearer /></Route>}
```
dpordomingo:
```suggestion
            {devmode && <Route path='/bearer'><Development.Bearer /></Route>}
```

src/js/Routes.jsx
```diff
@@ -18,13 +18,16 @@ import Stage from 'js/pages/pipeline/Stage';
 import Overview from 'js/pages/pipeline/Overview';
 import Login from 'js/pages/auth/Login';
 import Logout from 'js/pages/auth/Logout';
+import Bearer from 'js/components/development';
```
dpordomingo:
```suggestion
import Development from 'js/components/development';
```

src/js/components/development/index.jsx
```diff
@@ -32,5 +34,26 @@ export default {
...
+        return (
+          <Simple>
+            Not authenticated!
+          </Simple>
+        );
+      }
+
+      (async () => setToken(await getTokenSilently()))();
+
+      return <Simple><pre class="bearer-token">{token}</pre></Simple>;
```
dpordomingo: It must not be used `class`. Use `className` instead. :man_shrugging: 
```suggestion
      return <Simple><pre className="bearer-token">{token}</pre></Simple>;
```

vmarkovtsev: Hell why???

vmarkovtsev:
OK, it works now :+1: So My error was in the import syntax, and JS did not tell me that something was wrong. This is so JS :angry:

dpordomingo:
It was not a syntax error.
If you do: 
```jsx
import something from 'somewhere'
```
you're importing the default (no-named) export under `something`.

If you're using named exports:
```jsx
import {foo} from 'somewhere'
```
It will take the export named `foo`.

If you're using named exports you can also make an alias:
```jsx
import {foo as bar} from 'somewhere'
```

and mix both, importing the default one plus the named ones
```jsx
import something, {foo as bar} from 'somewhere'
```
And you'll be able to use `something` and `bar`.

Since `js/components/development/index.jsx` is exporting a default (no named)
```jsx
export default {
  foo: () => {}
}
```
and you did 
```jsx
import foo from 'js/components/development'
```
what you got under `foo` was the whole exported object, and that's exactly the error that React rised:
>_Error: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: object._

Translation:

>_I expected a function (a component to be rendered), but I got an object (`foo` is holding the whole default exported (no named) object)_

 so you should use instead:
```jsx
import SomethingNotMissleading from 'js/components/development'
<SomethingNotMissleading.foo>
```

vmarkovtsev:
Great explanation. :+1:

src/js/components/development/index.jsx
dpordomingo: because https://github.com/facebook/react/issues/13525#issuecomment-417818906
image

PR #91, opened 2020-03-03
Persist the current API schema

dpordomingo:
caused by https://github.com/athenianco/athenian-webapp/pull/75/files#r383942255

Once a new API schema is successfully used to autogenerate the API client, its sources will be stored into as 'src/js/services/api/openapi.v*.yaml' and the old ones will be removed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,8 @@ API_CLIENT_CODE_PATH := $(API_SERVICE_PATH)/openapi-client
 API_SPEC_PATH ?= $(API_SERVICE_PATH)/openapi.$(API_VERSION).yaml
 API_CLONE_PATH ?= /tmp/athenian-api
 
+OLD_API_SPEC := $(shell ls $(API_SERVICE_PATH)/openapi.*.yaml)
+
 OPENAPI_GENERATOR := node_modules/@openapitools/openapi-generator-cli/bin/openapi-generator
 
 $(API_SPEC_PATH):
@@ -33,7 +35,10 @@ $(API_SPEC_PATH):
 	rm -rf $(API_CLONE_PATH)
 
 .PHONY: api-generate
-api-generate: $(OPENAPI_GENERATOR) clean-openapi $(API_SPEC_PATH)
+api-generate: $(API_CLIENT_CODE_PATH)/index.js
+
+$(API_CLIENT_CODE_PATH)/index.js: $(OPENAPI_GENERATOR) $(API_SPEC_PATH)
+	$(MAKE) clean-openapi
 	node_modules/@openapitools/openapi-generator-cli/bin/openapi-generator generate \
 		--input-spec=$(API_SPEC_PATH) \
 		--output=$(API_CLIENT_CODE_PATH) \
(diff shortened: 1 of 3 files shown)


dpordomingo:
@vmarkovtsev since you're the API specs maintainer, and a `Make` expert as well, do you foresee any major issue with this change?

Makefile
```diff
@@ -33,6 +33,7 @@ $(API_SPEC_PATH):
 	rm -rf $(API_CLONE_PATH)
 
 .PHONY: api-generate
+api-generate: old_spec := $(shell ls -1 $(API_SERVICE_PATH)/openapi.*.yaml)
```
vmarkovtsev:
```suggestion
old_api_spec := $(shell ls $(API_SERVICE_PATH)/openapi.*.yaml)
```
(no need for the target on the left, renamed to `old_api_spec`, removed `-1`)

Makefile
```diff
@@ -41,6 +42,7 @@ api-generate: $(OPENAPI_GENERATOR) clean-openapi $(API_SPEC_PATH)
 		--ignore-file-override=$(API_SERVICE_PATH)/.openapi-generator-ignore \
 		--config=$(API_SERVICE_PATH)/.openapi-config.yml \
 		-DmodelTests=false,modelDocs=false,apiTests=false,apiDocs=false;
+	[ "$(old_spec)" != "$(API_SPEC_PATH)" ] && rm -f $(old_spec) || echo "used existing definition";
```
vmarkovtsev:
```suggestion
	[ "$(old_spec)" != "$(API_SPEC_PATH)" ] && rm -f $(old_spec)
```
(the file name can match but the timestamp can be newer if you edited it by hand, so not always "used existing definition").

Makefile
```diff
@@ -33,6 +33,7 @@ $(API_SPEC_PATH):
 	rm -rf $(API_CLONE_PATH)
 
 .PHONY: api-generate
+api-generate: old_spec := $(shell ls -1 $(API_SERVICE_PATH)/openapi.*.yaml)
 api-generate: $(OPENAPI_GENERATOR) clean-openapi $(API_SPEC_PATH)
```
vmarkovtsev: I'd suggest this:

1. Remove `.PHONY api-generate`
2.
```suggestion
api-generate: $(API_CLIENT_CODE_PATH)/index.js
$(API_CLIENT_CODE_PATH)/index.js: $(OPENAPI_GENERATOR) clean-openapi $(API_SPEC_PATH)
```

This is more idiomatic: if the file does not exist or is older than the spec or the spec name has changed, it gets regenerated.

vmarkovtsev:
BTW @dpordomingo now that the API spec is versioned independently from the server, we should probably switch to spec tags instead.

Makefile
dpordomingo: I like `2`, but I do not understand the purpose of `1`.
Why removing the `.PHONY api-generate` if `api-generate` is not a regular file to be generated by `make api-generate`?

dpordomingo:
@vmarkovtseva All your suggestions were added by d2c619e 

Since [`spec-1.0.5`](https://github.com/athenianco/athenian-api/releases/tag/spec-1.0.5) includes breaking changes, I'll bump to this new version in a new PR.

dpordomingo:
I'm merging this since all feedback from the review was addressed.
If you have any doubt/concern/suggestion about https://github.com/athenianco/athenian-webapp/pull/91#discussion_r387268321 let me know, please, and I'll answer/fix.

PR #93, opened 2020-03-03
Update API client to 1.0.5

dpordomingo:
depends on https://github.com/athenianco/athenian-webapp/pull/91

As suggested by https://github.com/athenianco/athenian-webapp/pull/91#issuecomment-594091171 it will be now used  spec tags instead, starting by [`spec-1.0.5`](https://github.com/athenianco/athenian-api/releases/tag/spec-1.0.5).

### minors:
- 01b7899 Upgrade openapi generator to 1.0.10-4.2.3


dpordomingo:
@vmarkovtsev [this](https://github.com/athenianco/athenian-webapp/pull/93/commits/25e547375da1edca21f69e12f6f78dd1c74aab30#diff-b67911656ef5d18c4ae36cb6741b7965) is what you requested by https://github.com/athenianco/athenian-webapp/pull/91#issuecomment-594091171

Makefile
```diff
@@ -16,31 +16,39 @@ $(MAKEFILE):
...
+	git -C $(API_CLONE_PATH) checkout $(API_VERSION_TAG_NAME) -- server/athenian/api/openapi/openapi.yaml
 	mv $(API_CLONE_PATH)/server/athenian/api/openapi/openapi.yaml $(API_SPEC_PATH)
 	rm -rf $(API_CLONE_PATH)
 
 .PHONY: api-generate
-api-generate: $(OPENAPI_GENERATOR) clean-openapi $(API_SPEC_PATH)
+api-generate: $(API_CLIENT_CODE_PATH)/index.js
+
+$(API_CLIENT_CODE_PATH)/index.js: $(OPENAPI_GENERATOR) $(API_SPEC_PATH)
+	$(MAKE) clean-openapi
```
vmarkovtsev: I wonder why you had to move it from the dependencies to a separate call.

dpordomingo: I found that when a `.PHONY` target (`clean-openapi`) is a pre is a prerequisite of a file target (`$(API_CLIENT_CODE_PATH)/index.js`), it is called **always** (which makes sense since the prerequisites of a file target need to be checked).

Then, every time `$(API_CLIENT_CODE_PATH)/index.js` is checked, `clean-openapi` is actually run, so `$(API_CLIENT_CODE_PATH)/index.js` is deleted as part of its pre-requisite checking, thus `$(API_CLIENT_CODE_PATH)/index.js` is always re-generated, no matter if it already existed and was newer than `$(API_SPEC_PATH)`.

Also, I'd say that `clean-openapi` should be only run **IF** `$(API_CLIENT_CODE_PATH)/index.js` is going to be generated, not as part of the process of checking its pre-requisites.

**TL;DR**
An advice I found at [clarkgrubb's makefile style guide](https://clarkgrubb.com/makefile-style-guide#toc7) was:
>_Phony targets should not be prerequisites of file targets._
>_The last rule is advised because a target with a phony target will always be executed._

PR #97, opened 2020-03-04
Centered alignment of the PR tabs

znegrin:
This task is related to the issue [#ENG-281](https://athenianco.atlassian.net/jira/software/projects/ENG/boards/4?assignee=5e57d3bc459a810c9af2098b&selectedIssue=ENG-281)

I fixed the alignment of the PR tabs and also changed the highlight color to the proper light orange:

image

### DIFF
--- a/src/sass/_overrides.scss
+++ b/src/sass/_overrides.scss
@@ -50,7 +50,7 @@ $font-size-sm:                $font-size-base * .75 !default;
 // Color variables
 
 $theme-colors: (
-        "primary": #FF7427,
+        "primary": #FFA008,
         "secondary": #8889A1,
         "success": #2FCC71,
         "danger": #EF3837,
--- a/src/js/components/pipeline/PullRequests.jsx
+++ b/src/js/components/pipeline/PullRequests.jsx
@@ -190,7 +190,7 @@ export default ({ data }) => {
         <>
             <div className="row mb-4">
                 <div className="col-12">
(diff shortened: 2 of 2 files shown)


src/sass/_overrides.scss
```diff
@@ -50,7 +50,7 @@ $font-size-sm:                $font-size-base * .75 !default;
 // Color variables
 
 $theme-colors: (
-        "primary": #FF7427,
+        "primary": #FFA008,
```
dpordomingo: We already have a company palette at [src/sass/_variables.scss](https://github.com/athenianco/athenian-webapp/blob/d14143e7b8bdccaf024296e885f76a90e303ffad/src/sass/_variables.scss)

What if we open a new PR in order to make this `_overrides.scss` file to use only colors from our palette at  `_variables.scss`?
E.g, I see that current `_overrides.scss` contains many colors already defined in our `_variables.scss`, but instead of using the existent variable, it is hardcoding the color.

znegrin: I suppose that you mean we should add the necessary colors as variables on the `_variables.scss` and then call them from `_overrides.scss` ?

dpordomingo: yes, it will also avoid the same color being outdated after changing it somewhere, but not there.

dpordomingo: If you want, we can address this problem in a new PR, because this one seems to be focused on the alignment of the PRs table.

PR #98, opened 2020-03-04
Fixing User Dropdown Styles

znegrin:
This task is based on the issue [#ENG-281](https://athenianco.atlassian.net/jira/software/projects/ENG/boards/4?assignee=5e57d3bc459a810c9af2098b&selectedIssue=ENG-281)

I fixed the styles on the User Dropdown to make it match the prototypes:

image

### DIFF
--- a/src/sass/components/_dropdowns.scss
+++ b/src/sass/components/_dropdowns.scss
@@ -24,7 +24,7 @@
     }
 
     .invitationCreator {
-        $copyButtonWidth: 5.5rem;
+        $copyButtonWidth: 7.5rem;
 
         .desc {
             $invitationDescMargin: map-get($spacers, 3);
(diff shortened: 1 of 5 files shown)


src/sass/components/_dropdowns.scss
```diff
@@ -32,7 +33,11 @@
         }
 
         button {
-            width: $copyButtonWidth;
+            width: 7.5rem;
```
dpordomingo: shouldn't it be updated the variable instead?

znegrin: You are right, I changed it. 
Thank you!

dpordomingo:
We have a conflict to be solved here.
Ping me when ready, and I'll merge.
thanks!

znegrin:
> We have a conflict to be solved here.
> Ping me when ready, and I'll merge.
> thanks!

@dpordomingo It is solved now. Thank you!

PR #106, opened 2020-03-10
Minors

dpordomingo:
blocks ____

This PR does not add visual changes.

Relevant changes:
- 3ccd914 Make sure that `make serve` does not modify `yarn.lock`; it should be manually upgraded using `yarn upgrade` instead.
- 73e30a2 `src/js/services/format::number` exports `percentage` and `round` helpers.

Other:
- ec44f02 Separate PR `participants` into `mergers` and `commentersReviewers`; rename `creators` as `authors`

Mocks:
- b40e721 In order to have data for the LeadTime, the `release` stage and `lead-time` have been mocked. It will be removed once the API starts sending them.

Minors:
- a6bfa76 Upgrade yarn dependencies
- e9ed9bf Cosmetics. Lint code

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -67,8 +67,7 @@ build:
 	CI=false yarn build
 
 .PHONY: serve
-serve:
-	yarn install
+serve: app-dependencies
 	yarn start
 
 .PHONY: test
@@ -83,6 +82,6 @@ dependencies: app-dependencies
 
 .PHONY: app-dependencies
 app-dependencies:
-	yarn install
+	yarn install --frozen-lockfile
 
 $(OPENAPI_GENERATOR): app-dependencies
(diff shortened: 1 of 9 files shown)


dpordomingo:
@znegrin @vmarkovtsev  After this PR is merged, `yarn.lock` should not be automatically changed in your regular workflow.
Please, let me know if it does because it would be an issue that I should solve.
Thanks!

Makefile
```diff
@@ -83,6 +82,6 @@ dependencies: app-dependencies
 
 .PHONY: app-dependencies
 app-dependencies:
-	yarn install
+	yarn install --frozen-lockfile
```
vmarkovtsev: I had a similar problem in Go, though Go mod does not have lock files but rather much more useless garbage. My solution was to exclude jt from the repo. Two years later, not a single issue appeared with that. Regarding JS, I agree that the lock file should be included in the repo.
Sorry, I just hate Go modules...

PR #107, opened 2020-03-10
Add Summary Metrics chart into Overview & Stages

dpordomingo:
This PR adds the Summary Charts into Overview and to all the pipeline stages.

- fef6c1a It is added a new chart: `FilledAreaChart`
- d8d6436 Add SummaryMetric component into Overview and each Stage; It's using already retrieved `prs`, `leadtime` and `stages` which are stored in `PRsContext` and `PipelineContext`. There is no data calculation in this commit, so it's used mocked data instead.
- ed5c3e5 Calculates the aggregations needed for the Summary Metrics following ￼[DesignDoc](https://docs.google.com/document/d/1DXLVz4L1Ag9sOUMAKdRRgVh1fk6peUrvVIp2xzeCRzQ)

image
image
image
image
image

### DIFF
--- a/docs/SummaryMetrics.md
+++ b/docs/SummaryMetrics.md
@@ -1,0 +1,69 @@
+# SummaryMetrics
+
+Design doc at https://docs.google.com/document/d/1DXLVz4L1Ag9sOUMAKdRRgVh1fk6peUrvVIp2xzeCRzQ
+
+## Common for all the stages
+
+- AVG:
+    requested separatelly
+- %:
+  ```
+  100 x Avg.stage / LeadTime
+  ```
+
+## WIP
+- Created PRs: Pull Requests Created
+  ```
+  prs.filter(from<pr.created<to).count
+  ```
+- Contributors: Number of PR creators of PRs created during the period selected
+  ```
+  prs.filter(from<pr.created<to).creators.distinct.count
+  ```
+- Repos: where commits have been pushed during the time period selected
+  ```
+  prs.filter(from<pr.created<to).repos.distinct.count
+  ```
+
+## Review
+- Reviewed PRs: Pull Requests Reviewed (PRs where a review has been submitted or a regular comment posted)
+  ```
+  prs.filter(pr.stage>=review).having(comments|reviews).count
+  ```
+- Reviewers: authors of one review or regular comment among the Reviewed PRs
+  ```
+  prs.filter(pr.stage>=review).participants.filter(p.status.oneOf(reviewer,commenter)).distinct.count
+  ```
+- Repos: where such comments have been posted
+  ```
+  prs.filter(pr.stage>=review).having(comments|reviews).repos.distinct.count
+  ```
+
+## Merge
+- Merged PRs:
+  ```
+  prs.filter(pr.merged=true).count
+  ```
+- Contributors: Number of “mergers”
+  ```
+  prs.filter(pr.merged=true).merger.distinct.count
+  ```
+- Repos: where such merges have happened
+  ```
+  prs.filter(pr.merged=true).repos.distinct.count
+  ```
+
+## Release
+- Released PRs: Pull Requests Released (PRs part of releases that happened during the time period selected)
+  ```
+  prs.filter(pr.stage=done&&pr.merged).count
+  ```
+- Releases:
+  ```
+  NOT-POSSIBLE; Should come from upcoming /releases endpoint
+  ```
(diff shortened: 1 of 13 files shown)


dpordomingo:
@znegrin you might be interested in reviewing the new Summary Metrics component once this is merged. I just used already defined colors and standard spacings. Feel free to make any suggestion you want.

@se7entyse7en You might be interested in reviewing only:
- fef6c1a It is added a new chart: `FilledAreaChart`
- ed5c3e5 Calculates the aggregations needed for the Summary Metrics following ￼[the DesignDoc](https://docs.google.com/document/d/1DXLVz4L1Ag9sOUMAKdRRgVh1fk6peUrvVIp2xzeCRzQ)

The other commit is mostly HTML, CSS plus a lot of wiring to move variables from A to B, which most probably will change once we change how the data is served from the API (instead of being calculated from the browser), and once we refactor React context ussage.

src/js/components/charts/FilledAreaChart.jsx
```diff
@@ -0,0 +1,47 @@
...
+      <FlexibleWidthXYPlot
+        height={height}
+        margin={{ left: 100 }}
+      >
+        <XAxis tickTotal={6}
+          tickFormat={dateFormat}
+        />
+        <YAxis tickTotal={3} tickFormat={dateTime.human} />
+        <HorizontalGridLines tickTotal={3} />
+        <VerticalGridLines tickTotal={6} />
```
se7entyse7en: Are the number ticks and the grid defined by product? Maybe at some point we would need to make this as variables.

docs/SummaryMetrics.md
```diff
@@ -0,0 +1,69 @@
+# SummaryMetrics
```
se7entyse7en: Thanks for this summary

src/js/components/pipeline/StageMetrics.jsx
```diff
@@ -32,6 +48,45 @@ export default ({ title, metrics }) => {
...
+      </div>
+    </div>
+  );
+};
+
+export const StageSummaryKPI = ({ data }) => {
+  return (
+    <div className="row">
+      {data.map((kpi, key) =>
+        <div className="col-md-6 mb-3" key={key}>
```
se7entyse7en: In theory, the number of rows should be `data.length / 2`. Maybe it worth `splice`-ing 2 by 2.

src/js/pages/pipeline/Pipeline.jsx
```diff
@@ -0,0 +1,175 @@
...
+            if (Array.isArray(processed)) {
+                processed.forEach(element => acc[element] = true);
+            } else {
+                acc[processed] = true;
+            }
+
+            return acc;
+        }, {})
+    );
+};
```
se7entyse7en: What about using `Set` and `flatMap`?

```suggestion
const distinct = (collection, extractor) => (new Set(x.flatMap((v) => extractor(v)))).keys();
```

docs/SummaryMetrics.md
dpordomingo: :)

src/js/components/charts/FilledAreaChart.jsx
dpordomingo: True, but I preferred to focus only on the main behavior, instead of in all its possible configurations and corner cases. Would you prefer if we make this configurable also in this PR?

src/js/components/pipeline/StageMetrics.jsx
dpordomingo: yes but... what benefit would we obtain from that?
I'd consider it only if there were any special design requirements for this layout.

src/js/pages/pipeline/Pipeline.jsx
dpordomingo: what a shame...
I should study the methods of `Array.prototype` :(

src/js/components/charts/FilledAreaChart.jsx
se7entyse7en: Not needed for this PR 👍

src/js/components/pipeline/StageMetrics.jsx
se7entyse7en: I just had a look [here](https://getbootstrap.com/docs/4.4/layout/grid/#column-wrapping). Since actually in this case they're all the same columns, it's actually easier as is.

PR #109, opened 2020-03-10
Sentry integration

se7entyse7en:
This PR adds the followings:
- configures `config.js` with the correct values to use Auth0 tenant for development and the corresponding Github App for development (no more need to use personal accoiunts),
- integrates Sentry by specifying `environment` and `release`:
    - the `release` is required in order to associate the sourcemaps to the minified js on Sentry,
    - when running in development, the release is not used since it would require pushing sourcemaps and it doesn't make sense given that in development mode errors are already clear,
    - for now the value used as release is simply a short version of the commit, in future we could consider using a real version,
    - the release value is set during the `build` step into a new file named `meta.js` that aims to contain all the metadata of the webapp.


se7entyse7en:
I forgot to install sentry-cli on travis, I'm on it.

se7entyse7en:
This depends on [this](https://github.com/athenianco/infrastructure/pull/61).

.travis.yml
```diff
@@ -30,7 +33,7 @@ jobs:
     - stage: build
       name: 'Frontend Build'
       script:
-        - make build
+        - RELEASE=${TRAVIS_COMMIT:0:7} make build
```
se7entyse7en: I've just realized that this should be done during the building of the docker image during the `deploy` stage. I'll update the PR.

se7entyse7en: Ah no ok, I think it's ok doing it in this step, probably just need to avoid pushing sourcemaps when not in `master` or `production` branch.

se7entyse7en: wdyt?

dpordomingo: I think this would be better in the deploy stage.

vmarkovtsev:
Is it possible to tag releases and assign semantic versions :pray:? I've recently had to check whether a certain commit was already released in prod (removal of `stages`), and instead of quickly looking at the version number running there I had to locate and go to the PR, copy the hash, then open the commit history, then search for that commit, then remember the position, then search in the same history for the commit currently running in prod, then compare the positions. Somewhat reminds me of developer tools vs `/bearer`...

se7entyse7en:
@vmarkovtsev I totally agree wrt semantic versioning, also for the purpose of deployment is a pain. But I remember that there was some discussion with @dpordomingo as well, but I totally forgot what was the point of the whole discussion. @dpordomingo can you remind me please?

vmarkovtsev:
Here: https://athenian.slack.com/archives/CQUFEUP5H/p1581962205023800

se7entyse7en:
> Here: https://athenian.slack.com/archives/CQUFEUP5H/p1581962205023800

It seems to me that in the end there was no disagreement on whether we need them. The only thing is that we would need to do the same as api that is always creating a tag since `master` could be deployed after any PR merged. I'll create a separate ticket for this.

se7entyse7en:
@dpordomingo moved to deploy stage, I'm gonna merge once CI green and check on staging

PR #115, opened 2020-03-12
Metrics groups

dpordomingo:
Add Metrics Group prototype into /prototypes/metrics-groups


dpordomingo:
@warenlg this is already at staging, so you can review the metrics groups at https://app-staging.athenian.co/prototypes/metrics-groups

warenlg:
Yes it looks great David, thanks for the ping!

PR #116, opened 2020-03-13
Better separation between the summary and the insights section

se7entyse7en:
The purpose of this PR is to reduce future conflicts. Since @dpordomingo you're going to add the `Insights` tab and I'm going to implement the charts inside that tab, I'd rather better split the summary block from the insights one.

So that I'll work on the Insights component and on the other hand you're free to move it when you'll implement the tab mechanism.

### DIFF
--- a/src/js/components/pipeline/StageMetrics.jsx
+++ b/src/js/components/pipeline/StageMetrics.jsx
@@ -13,7 +13,7 @@ import { palette } from 'js/res/palette';
 
 import FilledAreaChart from 'js/components/charts/FilledAreaChart';
 
-export default ({ conf, metrics, children }) => {
+export default ({ conf, children }) => {
   return (
     <div>
       <SummaryMetric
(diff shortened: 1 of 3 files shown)


src/js/components/pipeline/StageMetrics.jsx
```diff
@@ -32,22 +32,25 @@ export default ({ conf, metrics, children }) => {
...
-            }
-            insights={chart.insights}>
-          </Metric>;
-        })
-      }
     </div>
   );
 }
 
+export const Insights = ({metrics}) => (
```
se7entyse7en: This is just a copy-paste extracted from the upper component.

se7entyse7en:
@dpordomingo I started already working on the charts and I realized that it was better to PR this first to avoid conflicts later. I'll rebase my local branch for charts on top of this.

src/js/components/pipeline/StageMetrics.jsx
```diff
@@ -32,22 +32,25 @@ export default ({ conf, metrics, children }) => {
...
+    <>{
```
dpordomingo: I just realized that you used 4 spaces for tabs, but we're using 2 spaces.
Just saying so for your next prs ;)

PR #117, opened 2020-03-13
Fix config.js in order to use athenian development tenant

se7entyse7en:
This removes the need to use personal accounts for both GH and Auth0.

### DIFF
--- a/public/config.js
+++ b/public/config.js
@@ -1,16 +1,16 @@
 window.ENV = {
   auth: {
-    "domain": "yourtenant.auth0.com",
-    "clientId": "your-cliend-id",
+    "domain": "athenian-development.auth0.com",
+    "clientId": "oodpoYJUvWPyjjsrXjEsSgwE7PROni2r",
     "redirectUri": "http://localhost:3000/login/callback",
     "logoutRedirectUri": "http://localhost:3000",
-    "audience": "https://yourtenant.auth0.com/api/v2/"
+    "audience": "http://localhost:8080"
   },
   api: {
     "basePath": "http://localhost:8080/v1"
   },
   application: {
-    "githubAppUri": "https://github.com/apps/athenian-owl"
+    "githubAppUri": "https://github.com/apps/development-athenian-owl"
   },
   environment: "development"
 };


public/config.js
```diff
@@ -1,16 +1,16 @@
 window.ENV = {
   auth: {
-    "domain": "yourtenant.auth0.com",
-    "clientId": "your-cliend-id",
+    "domain": "athenian-development.auth0.com",
+    "clientId": "oodpoYJUvWPyjjsrXjEsSgwE7PROni2r",
     "redirectUri": "http://localhost:3000/login/callback",
     "logoutRedirectUri": "http://localhost:3000",
-    "audience": "https://yourtenant.auth0.com/api/v2/"
+    "audience": "http://localhost:8080"
```
dpordomingo: according to the previous value, shouldn't this use the new tenant?
```suggestion
    "audience": "https://athenian-development.auth0.com/api/v2/"
```

se7entyse7en: That value is correct already because it expects the api to be running locally at `localhost:8080` so also the name of the audience is `http://localhost:8080`.

se7entyse7en:
> I'm not sure how will this impact the whole experience regarding to how to get the `ATHENIAN_DEFAULT_USER` to be set in the API .env.
Since I already planned an update of the docs, could you please suggest how to get that id without accessing the Auth0 panel, and I'll add it into the docs when enhancing them?

Actually I didn't think about the default user. What I did was to log using my user and to hack the api db locally. I don't see much the point of using the default user, why don't we use our own user?

public/config.js
dpordomingo: then, I don't remember why did we need to define the `audience` as `https://<yourtenant>.auth0.com/api/v2/`. I remember that I had some problems when using `localhost:8080`, but since you already tested this, let's merge as it is.

dpordomingo:
If I remember properly, it was needed to run the API with proper `ATHENIAN_DEFAULT_USER`. But as I said, I'll go over that once I revisit the docs. Let's merge then.

@znegrin This was the change that I mentioned this morning. After merging this, you should be able to discard your local changes of `public/config.js`, stop ignoring its local changes, and use the ones from `master`. Let us know if it didn't work.

PR #119, opened 2020-03-16
Testing filter styles with a new library

znegrin:
This PR is related to the issue [ENG-291](https://athenianco.atlassian.net/browse/ENG-291).

I have been doing some research and testing the dropdown filters, first customizing the current [react-select](https://react-select.com/) library, and then replacing it with the [react-multiselect-checkboxes](https://www.npmjs.com/package/react-multiselect-checkboxes).

I manage to add the checkboxes (although it seems to not select them when all are selected by default):

image

I also manage to add a search field on top to replicate the behavior of the prototypes. I still couldn't add the buttons on the bottom, so I will maybe need help for that and also to [replace components](https://react-select.com/components#replacing-components), for example to add a "Select/Unselect All" checkmark on top, and to style the `control` area, like on the prototype:

image

I suppose this is something that needs evaluation by @dpordomingo and @warenlg, in terms of time and priority.


warenlg:
@dpordomingo any idea why the build is failing here?

dpordomingo:
I could not find the source code of the proposed library [`react-multiselect-checkboxes`](https://www.npmjs.com/package/react-multiselect-checkboxes) on GH, and considering only npm info, it seems to be not updated since more than 1y ago.

dpordomingo:
I could not find the source code of the proposed library [`react-multiselect-checkboxes`](https://www.npmjs.com/package/react-multiselect-checkboxes) on GH, and considering only npm info, it seems to be not updated since more than 1y ago.
Also, `react-multiselect-checkboxes` is using the older `react-select v.2`, so we'd miss all what's [new at  `react-select v.3`](https://github.com/JedWatson/react-select/issues/3585).
Maybe both are legit reasons because [it is currently used by only one other project](https://www.npmjs.com/browse/depended/react-multiselect-checkboxes).

Considering what said above, I'll try to reuse what was proposed here when adjusting the current filters.

se7entyse7en:
Indeed `react-multiselect-checkboxes` it's dead, but we can see how it's implemented and adapt it to use `react-sect v3` if it's not too much code.

PR #125, opened 2020-03-19
Integrate DateInterval calendar using "react-dates"

dpordomingo:
closes [[ENG-355]]

This PR implements the DateInterval filter using `airbnb/react-dates`

image

[ENG-355]: https://athenianco.atlassian.net/browse/ENG-355

### DIFF
--- a/src/js/components/ui/filters/DateInterval.jsx
+++ b/src/js/components/ui/filters/DateInterval.jsx
@@ -1,63 +1,115 @@
-import React, { useEffect, useState } from 'react';
 
-import AriaLabel from 'js/components/ui/AriaLabel';
+import React, { useState, useEffect } from 'react';
+import 'react-dates/initialize';
+import moment from 'moment';
+import { DateRangePicker } from 'react-dates';
+import { START_DATE, END_DATE } from 'react-dates/constants';
 
-import { dateTime } from 'js/services/format';
+const isInRange = (candidate, lower, upper) => lower.isBefore(candidate) && upper.isAfter(candidate);
 
-const max = (a, b) => a > b ? a : b;
-const min = (a, b) => a < b ? a : b;
-
-export default ({ id, name, className, dateIntervalLimits = { from: null, to: null }, onChange = () => { } }) => {
-    const fromMin = dateIntervalLimits.from || Date.now() - 365 * 24 * 60 * 60 * 1000;
-    const toMax = dateIntervalLimits.to || Date.now();
-    if (fromMin > toMax) {
-        throw new Error(`${id} date interval limits are wrong. ` +
-            `From:(${dateTime.ymd(fromMin)}) must be smaller than To:(${dateTime.ymd(toMax)})`);
+/**
+ * validateOrFixFn returns a function that calls resetFn if the passed dateInterval is empty, 
+ * or it will run fixFn over the field that could be missing in dateInterval; in both cases
+ * it will also return false; otherwise it will return true.
+ * @param {()=>{}} resetFn Function to be called in case the passed dateInterval is empty.
+ * @param {(string)=>{}} fixFn Function to be called in case the passed dateInterval lacks of any field.
+ * @returns {(<moment>)=>bolean} 
+ */
+const validateOrFixFn = (resetFn, fixFn) => dateInterval => {
+    if (!dateInterval.startDate && !dateInterval.endDate) {
+        resetFn();
+        return false;
     }
 
-    const [fromMaxState, setFromMaxState] = useState(toMax);
-    const [toMinState, setToMinState] = useState(fromMin);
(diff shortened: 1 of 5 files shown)


dpordomingo:
Styles to be adjusted by @znegrin in former PR.
@znegrin Feel free to propose any change you consider should be added here.

warenlg:
This already looks great! Thanks for the hard work David!

src/js/components/ui/filters/DateInterval.jsx
```diff
@@ -1,63 +1,115 @@
...
+        ) {
+            return;
+        };
+
+        setPrevDateIntervalState(dateIntervalState);
+        onChange({
+            from: dateIntervalState.startDate.valueOf(),
+            to: dateIntervalState.endDate.valueOf(),
+        });
+    }, [dateIntervalState, prevDateIntervalState, validateOrFix, focusedInputState, onChange]);
```
se7entyse7en: I don't understand why you need to do this. The library provides an `onClose` handler that you can pass and that is called only when you select correct dates and the calendar is closed: https://airbnb.io/react-dates/?path=/story/drp-calendar-props--with-onclose-handler.

You can check whether the date set is the same of the previous one during the `onClose`.

src/js/components/ui/filters/DateInterval.jsx
```diff
@@ -1,63 +1,115 @@
...
-        const newFromDateInterval = Date.parse(event.target.value) || fromMin;
-        setFromState(newFromDateInterval);
-        onChange({ from: newFromDateInterval, to: toState });
-    };
+export default ({
+    minDate = YEAR_AGO,
+    maxDate = EOD,
+    initialFrom = TWO_WEEKS_AGO,
+    initialTo = EOD,
+    onChange = dateInteval => console.log('APPLY', dateInteval)
```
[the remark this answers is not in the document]
dpordomingo: It's the default handle if none is passed to the calendar. But since it makes no sense to use that calendar without a change handler, it won't be used on a regular use case, no mater it is in staging nor prod

src/js/components/ui/filters/DateInterval.jsx
dpordomingo: That was my first attempt, but clear, is not triggering the close event. And that's only an example, for example, the calendar can be also closed with only the `startDate`.

src/js/components/ui/filters/DateInterval.jsx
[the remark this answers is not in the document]
se7entyse7en: Sorry, but I don't understand. This handler is not passed to the underlying date picker component.

src/js/components/ui/filters/DateInterval.jsx
se7entyse7en: Exactly, but that's the thing, isn't it? You're actually interested only when the interval selected Is complete. I guess that:
1. if the interval is fine, then update the filter and do the API call
2. if the interval is not correct or is the same of previous one, no API call

And you can achieve this just by using `onClose`.

src/js/components/ui/filters/DateInterval.jsx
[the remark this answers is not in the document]
se7entyse7en: Ah no ok, I got it now sorry. 👍 In this case though it would handy to wrap the function with a log in case of development, but not now of course.

src/js/components/ui/filters/DateInterval.jsx
dpordomingo: I don't understand your proposal.
We now handle the change event, and acting according to the final state of the filter:
- didn't change -> do nothing
- cleared -> set default
- incomplete -> ask for completion

Handling the close event does not provide any built-in help with that as far as I understand, and it would be needed to put the same logic in different event handlers, wouldn't it be?

se7entyse7en: I think that we could simplify it by avoiding to use the `useEffect`:
- `onClose`:
    - if didn't change, do nothing
    - if cleared, do nothing (to me it's more expected, but this is a Product thing that could be refined later)
    - if incomplete, do nothing as well

But not a blocker, let's merge this.

eiso:
Great work!

src/js/components/ui/filters/DateInterval.jsx
dpordomingo: What I tried to offer with my proposal as explained at https://github.com/athenianco/athenian-webapp/pull/125#discussion_r395633057, is a filter that is always showing the date interval that is actually filtering the data shown in the app.
My goal was to avoid the filter to be empty or in an incomplete state, so the user would not know what would be the time interval used.

Anyhow, let's merge and let @athenianco/product to validate this behavior and iterate considering also your suggestions, which are also fair.

PR #130, opened 2020-03-20
Count releases

dpordomingo:
image

In order to count releases for the given date interval, repos and devs, they're counted the distinct number of `release_url` of all the PRs returned for the same date interval, repos and devs, where `pr.release_url` is the URL of the earliest release that includes the PR.

### DIFF
--- a/docs/SummaryMetrics.md
+++ b/docs/SummaryMetrics.md
@@ -56,14 +56,13 @@ Design doc at https://docs.google.com/document/d/1DXLVz4L1Ag9sOUMAKdRRgVh1fk6peU
 ## Release
 - Released PRs: Pull Requests Released (PRs part of releases that happened during the time period selected)
   ```
-  prs.filter(pr.stage=done&&pr.merged).count
+  prs.filter(pr.release_url)
   ```
 - Releases:
   ```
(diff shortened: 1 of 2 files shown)


dpordomingo:
@warenlg @vmarkovtsev could you confirm that the assumption I did is correct?

warenlg:
Let's do it like this for now, so that we can move on to things with higher priority. 
However non of those 2 options are perfect: (1) rely on the `release_url` of the prs or (2) rely on the `released` datetime.
Because,
(1) is going to return a lot of releases that happen way after the date range selected (in case prs are created right before `time_to`)
(2) is more reliable regarding the date range filter but is not going to take into account the dev and repos filter.

In short, let's merge this PR and I'll raise a ticket to further improve the way we count releases.

dpordomingo:
`(1)` if a PR is created right before `time_to`, it's returned by `/filter/prs`, but is it going to include the `released_url` or `released` fields? It should not, but could you confirm it @vmarkovtsev ?
`(2)` we can not use `released` because it only contains the date when the PR was released, no we can not infer the actual release from there.

Just to clarify your concerns about releases not being filtered by the active filters:
>PRs being used to get the releases are **already filtered** by repo, contrib and date interval.

PR #135, opened 2020-03-23
Minor style adjustments to the date picker

znegrin:
This is related to the issue [#ENG-397](https://athenianco.atlassian.net/browse/ENG-397). I added some style adjustments to make the date picker more close to the prototypes.

image

### DIFF
--- a/src/js/components/ui/filters/DateInterval.jsx
+++ b/src/js/components/ui/filters/DateInterval.jsx
@@ -4,16 +4,18 @@ import 'react-dates/initialize';
 import moment from 'moment';
 import { DateRangePicker } from 'react-dates';
 import { START_DATE, END_DATE } from 'react-dates/constants';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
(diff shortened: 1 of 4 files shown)


znegrin:
I just made an update to remove the keyboard shortcuts panel:

image

I also was talking to @warenlg about the possibility of changing the behavior as right now the selected dates only apply when clicking outside the dropdown. Maybe an option to value is to add an apply button as we have on the prototypes. I am researching the documentation to see if there's an option to do something more intuitive than clicking outside to apply. Maybe @dpordomingo have also any ideas about this.

znegrin:
I found this: https://github.com/airbnb/react-dates/issues/1318 and it seems to be actually possible to add any rendering after the calendar, so maybe we can use this to add the cancel and apply buttons:

image

znegrin:
After talking with @warenlg, I added another commit with a custom info block adding the two buttons. It just need the event handlers to work properly, so it applies the date and closes the overlay when clicking on `Apply` instead of when clicking outside. 

@dpordomingo can you check it to see if it is not too time consuming to add this behavior? Thank you

image

PR #137, opened 2020-03-24
Implement accept/cancel buttons in the date filter

dpordomingo:
caused by https://github.com/athenianco/athenian-webapp/pull/135#issuecomment-602519292
required by [[ENG-355]]

This PR adds the proper behavior to the already existent `close` and `apply` data filter buttons; the behavior agreed with @warenlg is:

- click outside -> (keep current behavior) if the filter is valid, apply; otherwise, avoid closing.
- click in cancel -> closes the filter without applying any change.
- click in apply -> (enabled only if the filter is valid) applies the new filter.

If the filter is not valid (startDate and endDate are not both valid), the `apply` button is disabled as shown below:

image

[ENG-355]: https://athenianco.atlassian.net/browse/ENG-355


dpordomingo:
@znegrin you might be interested in changing the `disabled` effect of the `apply` button when it's not active.

PR #136, opened 2020-03-23
Add "Pull Request Size" insight chart

se7entyse7en:
## Screenshot

Screenshot 2020-03-23 at 19 24 54

PR #132, opened 2020-03-20
Add "Most Active Developers" insight chart with initial scaffolding

se7entyse7en:
Closes [ENG-371](https://athenianco.atlassian.net/browse/ENG-371).

This PR provides a framework for adding different charts for the insights sections of each stage. This framework lets the charts being defined in a declarative way where it's possible to define how data are calculated and how are displayed.

As the first chart to test this scaffolding, the "Most Active Developers" chart for the "Work In Progress" has been implemented.

All the work of this PR is confined in `js/components/insights`. Everything done outside is just to attach the insights sections to the webapp: [0f13e72](https://github.com/athenianco/athenian-webapp/commits?author=se7entyse7en).

`lodash` has ben used to make data plumbing easier and smoother since it provides a good set of data manipulation functions and the ability to pipe them.

## Screenshot of the tests chart

Screenshot 2020-03-20 at 18 14 15

This chart is not meant to be the final one.

PR #138, opened 2020-03-24
Insights charts api call

se7entyse7en:
Depends on: #132, #136.

Refactors the code in order to let each chart be able to query their own data.

Only relevant commit: https://github.com/athenianco/athenian-webapp/pull/138/commits/bfaa5e838773abe96e454ea650bd10e2f47381e0


se7entyse7en:
@dpordomingo rebased.

PR #139, opened 2020-03-24
Add "Wait time for 1st review" insight chart

se7entyse7en:
Closes [ENG-358](https://athenianco.atlassian.net/browse/ENG-358).
Depends on: #132, #136, #138.

## Screenshot

Screenshot 2020-03-24 at 19 42 58

### DIFF
--- a/src/js/components/insights/Helper.jsx
+++ b/src/js/components/insights/Helper.jsx
@@ -6,10 +6,15 @@ import _ from 'lodash';
 // TODO: Passing data should be removed in favor of letting each chart
 // retrieving its own data using the api.
 export const getInsights = async (stage, api, context, data) => {
-    const charts = {
+    const chartsGetterFn = {
         'work-in-progress': getWipCharts,
         'review': getReviewCharts
-    }[stage]();
+    }[stage];
+
+    let charts = [];
+    if (chartsGetterFn) {
+        charts = chartsGetterFn();
+    }
 
     return Promise.all(
         charts.map(
--- a/src/js/components/insights/Helper.jsx
+++ b/src/js/components/insights/Helper.jsx
@@ -1,5 +1,5 @@
-import { getCharts as getWipCharts } from 'js/components/insights/stages/WorkInProgress';
-import { getCharts as getReviewCharts } from 'js/components/insights/stages/Review';
+import getWipCharts from 'js/components/insights/stages/WorkInProgress';
+import getReviewCharts from 'js/components/insights/stages/Review';
 
 import _ from 'lodash';
 
@@ -11,10 +11,7 @@ export const getInsights = async (stage, api, context, data) => {
         'review': getReviewCharts
     }[stage];
 
-    let charts = [];
-    if (chartsGetterFn) {
-        charts = chartsGetterFn();
-    }
+    const charts = chartsGetterFn ? chartsGetterFn() : [];
 
     return Promise.all(
         charts.map(
--- a/src/js/components/insights/Helper.jsx
+++ b/src/js/components/insights/Helper.jsx
@@ -1,19 +1,22 @@
-import { getInsights as getInsightsWorkInProgress } from 'js/components/insights/stages/WorkInProgress';
-import { getInsights as getInsightsReview } from 'js/components/insights/stages/Review';
+import { getCharts as getWipCharts } from 'js/components/insights/stages/WorkInProgress';
+import { getCharts as getReviewCharts } from 'js/components/insights/stages/Review';
 
 import _ from 'lodash';
 
 // TODO: Passing data should be removed in favor of letting each chart
 // retrieving its own data using the api.
-export const getInsights = (stage, api, context, data) => {
-    const fn = {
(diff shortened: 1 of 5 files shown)


eiso:
You're on fire. Nice work!

dpordomingo:
I did checkout to this latest PR, and something unexpected happened.
Are you maybe using any local data that I should also use to reproduce this?
image

se7entyse7en:
@dpordomingo thanks for noticing, I forgot to handle non-implemented insights section yet during the last refactor. Fixed here: https://github.com/athenianco/athenian-webapp/pull/139/commits/1f8a8d3319240aaccf6f6ba24373be602222cb49.

warenlg:
Test comment to see how fast we are at updating staging

src/js/components/insights/stages/Review.jsx
```diff
@@ -1,13 +1,167 @@
...
 import _ from 'lodash';
 
 export const getCharts = () => [
+    waitTimeFirstReview,
     pullRequestSize
 ];
 
+const waitTimeFirstReview = {
+    fetcher: async (api, context, data) => {
+        const metricIDs = new MetricID();
```
dpordomingo: We're trying to perform the API calls from the `services/api`. I think that moving `fetchChartsData` and `fetchKPIsData` there, and not depending here on `js/services/api/openapi-client`, would be also better in this case.

se7entyse7en:
@warenlg updated screenshot with our own data locally

Screenshot 2020-03-26 at 12 29 10

src/js/components/insights/charts/library/TimeSeries.jsx
```diff
@@ -0,0 +1,87 @@
...
+          }))
+          .value();
+
+    const tickValues = _(formattedData)
+          .map(v => v.x.getTime())
+          .value();
+
+    return (
+        <FlexibleWidthXYPlot height={300} margin={{ left: 100, right: 30 }}>
+          <VerticalGridLines tickValues={tickValues} />
```
dpordomingo: Do we really want to show as many vertical grid lines as data points?
Because this is not happening in all our charts, and it could be also a good idea in this kind of charts. wdyt @warenlg ?

example:
image

warenlg:
I'm fine with the chart.
But about the KPI showing the proportion of the Lead Time, are you sure you're making x100 since we are showing a percentage?

src/js/components/insights/charts/library/TimeSeries.jsx
se7entyse7en: I agree that we need to define and decide how all these things should appear: ticks frequency, grids, etc.. Let's keep this in mind for next iteration. It will be easier to define once we have a view as a whole.

src/js/components/insights/stages/Review.jsx
se7entyse7en: This is another of those things that I'll move separate. This is related to the thing we discussed about having a separate component that handles also caching and things like that. For now I wanted to have each chart have maximum flexibility as I don't know what kind of calls will be needed. I don't like this part either, but again I'm not aware on how to refactor this until we have a full view of the needs.

src/js/components/insights/Helper.jsx
```diff
@@ -6,10 +6,15 @@ import _ from 'lodash';
 // TODO: Passing data should be removed in favor of letting each chart
 // retrieving its own data using the api.
 export const getInsights = async (stage, api, context, data) => {
-    const charts = {
+    const chartsGetterFn = {
```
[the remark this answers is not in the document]
se7entyse7en: simplified the current version, but without using `switch` to avoid sugar code.

se7entyse7en:
@dpordomingo fixed the `unit` passed as a single string, made default exports and simplified `charts` fech functions call. Gonna merge after CI.

src/js/components/insights/stages/Review.jsx
dpordomingo: But moving (copy-pasting) it there is not incompatible with what we planned, it's only about consistency with what we already have.

se7entyse7en:
@warenlg sorry I forgot your comment, indeed I forgot x100 😂

se7entyse7en:
@warenlg ah no sorry, I looked at the flow ratio, the calculation of the proportion is fine.

src/js/components/insights/Helper.jsx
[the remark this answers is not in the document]
dpordomingo: is a switch code "sugar code"?
Imo it's more idiomatic than the associative array and an if, but ¯\_(ツ)_/¯  ok, not blocker :)

se7entyse7en:
@dpordomingo ok, gonna address after these two remaining PRs to at least move them already in `services`.

warenlg:
But it would mean that the Lead Time against what you divide is 5,000 hours...

se7entyse7en:
@warenlg okay stupid me, I was dividing the proportion by 3600 🤦‍♂

se7entyse7en:
😅

PR #141, opened 2020-03-25
Add "Pull Request Ratio Flow" insight chart

se7entyse7en:
Closes [#362](https://athenianco.atlassian.net/browse/ENG-362).
Depends on: #132, #136, #138, #139.

## Screenshot
Screenshot 2020-03-25 at 16 43 25

### DIFF
--- a/src/js/components/insights/stages/WorkInProgress.jsx
+++ b/src/js/components/insights/stages/WorkInProgress.jsx
@@ -1,11 +1,139 @@
 import { SimpleKPI } from 'js/components/insights/KPI';
+import TimeSeries from 'js/components/insights/charts/library/TimeSeries';
 import HorizontalBarChart from 'js/components/insights/charts/library/HorizontalBarChart';
 
+import { PullRequestMetricsRequest } from 'js/services/api/openapi-client';
+
+import ForSet from 'js/services/api/openapi-client/model/ForSet';
+import MetricID from 'js/services/api/openapi-client/model/MetricID';
+
+import { dateTime, github } from 'js/services/format';
+
+import moment from 'moment';
 import _ from 'lodash';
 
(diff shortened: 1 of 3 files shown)


src/js/components/insights/stages/WorkInProgress.jsx
```diff
@@ -1,12 +1,147 @@
...
 
 export const getCharts = () => [
-    mostActiveDevs
+    mostActiveDevs,
+    prRatioFlow
 ];
 
+const prRatioFlow = {
+    fetcher: async (api, context, data) => {
+        const metricIDs = new MetricID();
```
dpordomingo: the very concern I already have at https://github.com/athenianco/athenian-webapp/pull/139#discussion_r398493164

se7entyse7en:
@warenlg 

Screenshot 2020-03-26 at 13 39 57

warenlg:
Very nice chart!! we clearly see the weekends!
However, the 2nd KPI is suspicious to me: how can we have an avg PR ratio flow for Mondays of 3 when the max seems to be 2 on Feb 14th?

se7entyse7en:
Good point. The reason is this:

Screenshot 2020-03-26 at 13 53 31
Screenshot 2020-03-26 at 13 53 37

So basically `Monday` in the chart do not appear because `pr-flow` is `0` since denominator is `0`. But the overall KPI what does is `sum(opened) / sum(closed)` by day, so for `monday` is `3` even if the single datapoints are all zeros.

Do you think is better to change it to the average of the `pr-flow`?

se7entyse7en:
This has been solved on Slack. Each datapoint is calculated as `(opened || 1) / (closed || 1)` and the for a day we compute the average of the datapoints.

PR #144, opened 2020-03-26
Some cleanup

se7entyse7en:
Cleaned up things:
- make all API calls from `services/api`,
- remove warning related to insights charts,
- remove functions related to sample charts,
- found a way to properly inject `svg` tags required inside a react-vis plot. (This will be useful to refactor other components, but now now, probably once I'll deal with them again.)

### DIFF
--- a/src/js/services/api/index.js
+++ b/src/js/services/api/index.js
@@ -239,3 +239,24 @@ export const getSampleCharts = stage => {
 
   return new Promise(resolve => window.setTimeout(resolve(chartsSampleData), getRand(100, 1000)));
 };
+
+
+export const fetchPRsMetrics = async (
+    api, accountID,
+    granularity,
+    dateInterval,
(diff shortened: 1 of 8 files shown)


src/js/services/api/index.js
```diff
@@ -239,3 +239,27 @@ export const getSampleCharts = stage => {
...
+    granularity,
+    dateInterval,
+    metrics = [],
+    filter = { repositories: [], developers: [] }
+) => {
+    const metricIDs = new PullRequestMetricID();
+    const forset = ForSet.constructFromObject({
+        repositories: filter.repositories,
+        developers: filter.developers
+    });
```
dpordomingo: Why not?
```suggestion
    const forset = ForSet.constructFromObject(filter);
```

se7entyse7en: done

PR #159, opened 2020-03-31
Replacing logos with the definitive versions

znegrin:
This is based on the task [#ENG-458](https://athenianco.atlassian.net/browse/ENG-458). I replaced all the logos and icons with the definitive versions. I also changed the page title to just `Athenian`, as @eiso requested.

image

image


znegrin:
Hi @dpordomingo, I just made the requested changes.

In terms of resolution, all replaced icons are in the same resolution of the previous ones. In the case of the `favicon` I guess that is the `.ico` version which always has a lower resolution, but I can see the icon on my HD screen and it looks perfectly sharp as it probably takes the higher resolution image.

PR #162, opened 2020-03-31
Add basic overlay functionality to FilledAreaChart

se7entyse7en:
Adds basic overlay functionality. The styling is out of the scope of this PR.

## Screenshot

Screenshot 2020-03-31 at 17 28 27

PR #164, opened 2020-03-31
Tooltip component

se7entyse7en:
Creates a `Tooltip` component. The styling is out of the scope of this PR.

### DIFF
--- a/src/js/components/charts/Tooltip.jsx
+++ b/src/js/components/charts/Tooltip.jsx
@@ -1,0 +1,46 @@
+import React from 'react';
+
+import _ from 'lodash';
+
+import {
+    Hint
+} from 'react-vis';
+
+export default ({value, ...props}) => {
+    // TODO: This needs to be styled.
+    // `value` is a flat key-value object
+    return (
+        <Hint {...props} value={value}>
+          <div style={{background: 'black'}}>
+            <h4>Value of hint</h4>
+
+            {
+                _(value).map((v, k) => {
+                    // How each key-value is rendered
+                    return <p key={k}>{k}: {v.toString()}</p>;
+                }).value()
+            }
+
+          </div>
+        </Hint>
+    );
+};
+
+export const onValueChange = (datapoint, eventType, current, setCurrent) => {
+    console.log(`START: ${eventType}`);
+
+    if (!current ||
+        (current && (datapoint.x.toString() !== current.x.toString() ||
+                          datapoint.y !== current.y))) {
+        console.log('setting datapoint');
+        setCurrent(datapoint);
+    }
+
+    console.log(`END: ${eventType}`);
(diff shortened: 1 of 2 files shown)


se7entyse7en:
@znegrin related to what @warenlg mentioned, this PR isolates more the component to be style for the tooltips.

See [here](https://github.com/athenianco/athenian-webapp/pull/164/files#diff-5d2e4b9a3401825c053f7b5c6a195b52R9).

If you wanna see it in action, this is already attached to the 5 summary metrics charts as showed [here](https://github.com/athenianco/athenian-webapp/pull/162#issue-396413847).

src/js/components/charts/Tooltip.jsx
```diff
@@ -0,0 +1,46 @@
...
+export default ({value, ...props}) => {
+    // TODO: This needs to be styled.
+    // `value` is a flat key-value object
+    return (
+        <Hint {...props} value={value}>
+          <div style={{background: 'black'}}>
+            <h4>Value of hint</h4>
+
+            {
+                _(value).map((v, k) => {
```
se7entyse7en: @znegrin @warenlg for now this is very dummy and just creates `p` tag for each key-value.

src/js/components/charts/Tooltip.jsx
```diff
@@ -0,0 +1,46 @@
...
+
+export const onValueChange = (datapoint, eventType, current, setCurrent) => {
+    console.log(`START: ${eventType}`);
+
+    if (!current ||
+        (current && (datapoint.x.toString() !== current.x.toString() ||
+                          datapoint.y !== current.y))) {
+        console.log('setting datapoint');
+        setCurrent(datapoint);
+    }
```
dpordomingo: `!current || current` is redundant, also, I'm not sure why is it needed the `toString()`; I'd purpose the more simple alternative:
```suggestion
  if (current && datapoint.x === current.x && datapoint.y === current.y) {
    return
  }
  
  console.log('setting datapoint');
  setCurrent(datapoint);
```

src/js/components/charts/Tooltip.jsx
```diff
@@ -0,0 +1,46 @@
...
+
+    console.log(`END: ${eventType}`);
+};
+
+export const onValueReset = (datapoint, eventType, current, setCurrent) => {
```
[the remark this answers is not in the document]
se7entyse7en: for now that's the only thing + logging, but I'd rather keep bounded here together the logic for the tooltip.

src/js/components/charts/Tooltip.jsx
se7entyse7en: the `toString` was necessary because wasn't working for dates. btw this has been fixed already in next PR

PR #170, opened 2020-04-01
Rendering improvements and stabilization

se7entyse7en:
Closes [ENG-463](https://athenianco.atlassian.net/browse/ENG-463).

This PR refactors the `Filters` component in order to have a cleaner handling of the state changes and to expose with the `FilterContext` a proper `ready` props. That `ready` props is then used by `Stage` component to avoid triggering API calls if filters are not ready.

This PR also adds proper state management for the data of the insights chart for each stage. Those data are refreshed with an API call only if filters change or if no data has been stored yet.

The `Insights` component now is aware of whether the data are being loaded, this makes possible to show a loading message or to improve it later using a spinner.

The `Stage` component could be simplified in the future, in particular, wrt the callback of the `useEffect`. Some notes have been added and logging has been left for easier debugging.

### DIFF
--- a/src/js/components/insights/Insights.jsx
+++ b/src/js/components/insights/Insights.jsx
@@ -2,6 +2,12 @@ import React from 'react';
 
 import Box from 'js/components/insights/Box';
 
-export default ({insights}) => (
-    <>{insights.map((ins, i) => <Box meta={ins.meta} content={ins.content} key={i} />)}</>
-);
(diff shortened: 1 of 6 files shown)


se7entyse7en:
@dpordomingo this is a big change, you might be interested in reviewing this, and you'll probably spot some things to improve.

PR #174, opened 2020-04-02
Fix for typos and copy

rpau:
This patch fixes:
- The text style of some visual components to lowercase
- The copy to ensure that is more readable for humans and less "developer" style.
- Small typo

NOJIRA

### DIFF
--- a/src/js/components/insights/stages/review/pullRequestSize.jsx
+++ b/src/js/components/insights/stages/review/pullRequestSize.jsx
@@ -47,11 +47,11 @@ const pullRequestSize = {
         grouper: 'reviewRequested',
         groups: {
             false: {
-                title: 'Waiting Review',
+                title: 'Waiting review',
                 color: '#FFC507'
             },
             true: {
-                title: 'Review Submitted',
+                title: 'Review submitted',
                 color: '#23C2C7'
             }
         }
@@ -66,7 +66,7 @@ const pullRequestSize = {
                     chart: {
                         component: BubbleChart,
                         params: {
-                            title: 'Number of Pull Requests created',
+                            title: 'Number of created pull requests',
                             data: computed.chartData,
                             extra: {
                                 grouper: computed.grouper,
@@ -82,15 +82,15 @@ const pullRequestSize = {
                     },
                     kpis: [
                         {
-                            title: {text: 'Total Number of Files', bold: true},
-                            subtitle: {text: 'Waiting for Review'},
+                            title: {text: 'Total number of files', bold: true},
+                            subtitle: {text: 'Waiting for review'},
                             component: SimpleKPI,
                             params: {
                                 value: computed.totalFiles
                             }
                         },
                         {
-                            title: {text: 'Average Pull Request Size', bold: true},
+                            title: {text: 'Average pull request size', bold: true},
                             component: MultiKPI,
                             params: [
                                 {
--- a/src/js/components/pipeline/MainMetrics.jsx
+++ b/src/js/components/pipeline/MainMetrics.jsx
@@ -30,12 +30,12 @@ export default ({
             <div className="col-md-3">
                 <MainMetric
                     title="Pull requests" value={createdPRsAvg} variation={createdPRsVariation}
-                    hint="Number of created pull requests"
+                    hint="Number of pull requests processed through the delivery pipeline"
                 />
             </div>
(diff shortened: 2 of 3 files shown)


src/js/components/insights/stages/review/pullRequestSize.jsx
```diff
@@ -47,11 +47,11 @@ const pullRequestSize = {
         grouper: 'reviewRequested',
         groups: {
             false: {
-                title: 'Waiting Review',
+                title: 'Waiting review',
```
se7entyse7en, line 50: Regarding the copies, I think that all the copies in `js/components/insights` are in "Title"-case, if we wanna change it, then we'll need to change it in each chart. What I'd do is to write them with normal case and then have a layer that works with them consistently.

src/js/components/pipeline/MainMetrics.jsx
```diff
@@ -16,26 +16,26 @@ export default ({
...
-                    hint="The sum of each stage average time from start to finish"
+                    hint="The sum of the average time required in each development stage"
                     negativeIsBetter
                 />
             </div>
             <div className="col-md-3">
                 <MainMetric
                     title="Pull requests" value={createdPRsAvg} variation={createdPRsVariation}
-                    hint="Number of Pull Requests that hav been created within the active filters"
+                    hint="Number of created pull requests"
```
warenlg, line 33: This one was wrong before as well. It is the "Number of pull requests processed through the delivery pipeline"

src/js/components/pipeline/MainMetrics.jsx
```diff
@@ -16,26 +16,26 @@ export default ({
...
                 />
             </div>
             <div className="col-md-3">
                 <MainMetric title="Contributors" value={contribsAvg} variation={contribsVariation}
-                    hint="Number of people participating in the PRs within the active filters"
+                    hint="Number of people participating in pull requests"
```
warenlg, line 38: I would be more specific here with something like "...people participating in the (analyzed) pull requests"

src/js/components/pipeline/MainMetrics.jsx
```diff
@@ -16,26 +16,26 @@ export default ({
...
-                    hint="Average of time elapsed between the creation of the 1st commit in the Pull Requests within the active filters, and the code being used in production"
+                    hint="Elapsed time between the creation of the 1st commit in a pull request and the code being used in production"
                     negativeIsBetter
                 />
             </div>
             <div className="col-md-3">
                 <MainMetric
                     title="Cycle time" value={cycleTimeAvg && dateTime.human(cycleTimeAvg)} variation={cycleTimeVariation}
-                    hint="The sum of each stage average time from start to finish"
+                    hint="The sum of the average time required in each development stage"
```
warenlg, line 26:
```suggestion
                    hint="Sum of the average time required in each development stage"
```

src/js/components/pipeline/MainMetrics.jsx
warenlg: The number of created PRs is lower and displayed among the WIP summary metrics.

src/js/components/pipeline/MainMetrics.jsx
rpau: I removed the `active filters` part because it seemed redundant/obvious for me and makes the message longer. Do you think that it could create confusion?

src/js/components/insights/stages/review/pullRequestSize.jsx
rpau: I agree, but the titles were inconsistent. For example: `Number of Pull Requests that hav been created within the active filters`. 

I'll review everything in the insights folder to make it consistent. In order to reduce the scope of this PR, should I create a ticket for that layer?

rpau: There are not more hardcoded messages in that folder.

src/js/components/pipeline/MainMetrics.jsx
warenlg: The `active filters` should be removed for sure.
My suggestion was only to add "the" or even "the analyzed/processed" to make it clear that the contributors displayed are related to the current set of PRs filtered upstream.

rpau: So them I think that your requests have been addressed

PR #176, opened 2020-04-02
Fix variation 1st review kpi

dpordomingo:
required by [[ENG-433]]

`Average Waiting Time for 1st review` and `Proportion of the Cycle Time` KPIs of `Wait Time for 1st Review` chart, seems to be good when decreasing, and bad if increasing, so this PR changes badge color following that rule.

image

[ENG-433]: https://athenianco.atlassian.net/browse/ENG-433


eiso:
Nice work. Thank you!

PR #186, opened 2020-04-06
Implement 'User Icon' tooltip

dpordomingo:
`react-vis` does not support the native `<Hint>` for axis, so it was implemented, as a workaround, a text label next to the user image.
- If it is approved, it should be styled using SVG native elements to match our design for `User Icon` tooltip
- Alternative: investigate if it can be attached a javascript handler to `<svg:image:onHover>`, and use it to handle the `react-vis` `<Hint>` passing the `(x,y)` position of the triggered event.

To be used in `YAxis` of the following charts:
- Most Active Developer
- Review Activity :: ReviewComments/Reviews per reviewer

image

### DIFF
--- a/src/js/components/insights/charts/library/HorizontalBarChart.jsx
+++ b/src/js/components/insights/charts/library/HorizontalBarChart.jsx
@@ -60,21 +60,6 @@ const HorizontalBarChart = ({ title, data, extra }) => {
            }}/> : null
           }
 
-          {extra && extra.yAxis && extra.yAxis.imageMapping ?
-           <YAxis tickFormat={
-               (value) => (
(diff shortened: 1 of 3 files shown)


dpordomingo:
@se7entyse7en This acts as a workaround as described in the PR description. wdyt about:
- investigating the described alternative,
- or using this approach, and style the tooltip using svg elements

and/or postpone the decision after the release.

@warenlg would be the current state of the tooltip valid for releasing the MVP, or is it a no-go option?
image

warenlg:
Thank you for the PR @dpordomingo. For now, the tooltip is fine as it gives the required information. Knowing it's not straightforward, we'll raise another ticket later to style it better.

PR #188, opened 2020-04-06
Smart granularity

se7entyse7en:
Adds a smart granularity calculation for summary charts based on the following logic:

- <= 21 days => 'day'
- <= 90 days => 'week'
- else => 'month'

This PR also fixes the grid line by making it match with the x-axis labels.

## Screenshot

Screenshot 2020-04-06 at 12 44 11


se7entyse7en:
> You could even increase the day-week transition threshold to 21 or 28 days

I was afraid of the opposite, that is a chart with 21 data points, but I'm gonna try it 👍

se7entyse7en:
Ok 21 datapoints are still fine 👍

PR #190, opened 2020-04-06
Summary chart y values

se7entyse7en:
Fixes y-values in Summary Charts and average.

## Screenshot 

Screenshot 2020-04-06 at 13 12 21

### DIFF
--- a/src/js/pages/pipeline/Overview.jsx
+++ b/src/js/pages/pipeline/Overview.jsx
@@ -29,7 +29,7 @@ export default () => {
                     name={leadtimeContext.stageName}
                     metric={leadtimeContext.metric}
                     config={{
-                        average: leadtimeContext.avg
+                        average: leadtimeContext.avg / 1000
                     }}
                   />;
     const kpi = stagesContext.map((stage, i) => stage.overallProportion ? (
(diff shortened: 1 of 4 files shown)


se7entyse7en:
> As a side note, I'm not sure how you calculate the avg horizontal line but ideally the value should match the time displayed on the left in the summary metrics. Thus it should be the avg over all the points excluding the one on the left that we artificially added.

I didn' change the calculation of the avg other than fixing the unit (seconds vs millisecond).

se7entyse7en:
@warenlg merging this and will fix average issue separately.

src/js/pages/pipeline/Overview.jsx
```diff
@@ -29,7 +29,7 @@ export default () => {
                     name={leadtimeContext.stageName}
                     metric={leadtimeContext.metric}
                     config={{
-                        average: leadtimeContext.avg
+                        average: leadtimeContext.avg / 1000
```
dpordomingo, line 32: This is unexpected for me... here, we're receiving `ms` from `leadtimeContext.avg`, and passing `s` to the `SummaryChart`.
And then in the `SummaryChart` we receive `s` (because above), and convert them back to `ms`.
I guess I'm missing something.

se7entyse7en: This is exactly what's happening, let me clarify the rationale. The chart now is partially "standalone" in the sense that for the data it calls the API by itself, but the average is still provided externally the way it was before (from the data of the initial call). The self-API call returns data in seconds, so I implicitly made the interface of the chart to work with seconds as it was weird to have half data in seconds and in ms. Once all the data are plumbed into the charts in seconds, for formatting they're converted into ms because the formatting function interface requires ms.

PR #194, opened 2020-04-06
CI tests in docker container

se7entyse7en:
This PR makes tests and build run inside docker container itself:
- makes both and tests more reliable as they run exactly in the same environment
- makes CI more stable, for example, today Travis was failing continuously in downloading node 13 using nvm, this is not more required since everything is done inside containers.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,4 @@
-language: node_js
-node_js:
-  - '13'
+language: minimal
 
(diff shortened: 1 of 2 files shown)


se7entyse7en:
Merging this as it is blocking deploys.

dpordomingo:
great idea!

PR #196, opened 2020-04-07
Summary charts values formatting

se7entyse7en:
This PR adds dynamic y-label unit selection for summary charts depending on the maximum value. This also adds dynamic unit selection for each individual value for the tooltips.

## Example gif

dynamic-units

### DIFF
--- a/src/js/services/format.js
+++ b/src/js/services/format.js
@@ -13,6 +13,42 @@ const WEEK = 7 * DAY;
 const MONTH = 30 * DAY;
 const YEAR = 365 * DAY;
 
+const roundDecimals = (num, decimals) => Math.round((num + Number.EPSILON) * 10**decimals) / 10**decimals;
+
+export const getBestTimeUnit = (milliseconds) => {
+    const units = [
+        [SECOND, 'secs'],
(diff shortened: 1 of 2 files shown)


warenlg:
Also, currently, the points where the value is 0 sec is because we don't have data for that point, right?

se7entyse7en:
> Also, currently, the points where the value is 0 sec is because we don't have data for that point, right?

yup, that's hardcoded. I thought that `secs` is the "nicest" unit for `0`.

src/js/services/format.js
```diff
@@ -13,6 +13,42 @@ const WEEK = 7 * DAY;
 const MONTH = 30 * DAY;
 const YEAR = 365 * DAY;
 
+const roundDecimals = (num, decimals) => Math.round((num + Number.EPSILON) * 10**decimals) / 10**decimals;
```
dpordomingo, line 16: why not enhancing current [`round`](https://github.com/athenianco/athenian-webapp/blob/master/src/js/services/format.js#L118) instead of creating a new one?

se7entyse7en: because I completely missed it 😛

PR #198, opened 2020-04-07
Fix calculation of main metrics and variations

se7entyse7en:
Closes [ENG-478](https://athenianco.atlassian.net/browse/ENG-478).

This PR fixes the calculation of the main metrics and uses the `all` granularity to make queries more lightweight.

### DIFF
--- a/src/js/services/api/index.js
+++ b/src/js/services/api/index.js
@@ -84,84 +84,50 @@ export const getContributors = (token, userAccount, from, to, repos) => {
     .then(contribs => contribs.map(c => c.login));
 };
 
-export const getMetrics = (api, accountId, dateInterval, repos, contributors) => {
-  const metrics = ['lead-time', 'cycle-time', 'wip-time', 'review-time', 'merging-time', 'release-time'];
-  const result = {};
-  metrics.forEach(metric => {
-    result[metric] = {
-      data: [],
-      avg: 0,
-      variation: 0,
-      aux: {
-        prevInterval: { sum: 0, count: 0 },
-        mainInterval: { sum: 0, count: 0 },
-      }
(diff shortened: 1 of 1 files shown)


src/js/services/api/index.js
```diff
@@ -84,84 +84,50 @@ export const getContributors = (token, userAccount, from, to, repos) => {
...
+    const result = {};
+
+    _(metrics)
+        .forEach((m, index) => {
+            const prevAvg = dateTime.milliseconds(
+                prevResult.calculated[0].values[0].values[index]) || 0;
+            const currAvg = dateTime.milliseconds(
+                currResult.calculated[0].values[0].values[index]) || 0;
+
+            const variation = prevAvg > 0 ? (currAvg - prevAvg) * 100 / prevAvg : 0;
```
eiso, line 112: Shouldn't the `*100`, not sure of the precedence order in Javascript but in math, you would do it as: `(currAvg - prevAvg) / prevAvg * 100`

eiso: Results will be the same, so feel free to ignore this.

src/js/services/api/index.js
```diff
@@ -84,84 +84,50 @@ export const getContributors = (token, userAccount, from, to, repos) => {
...
+    const metrics = ['lead-time', 'cycle-time', 'wip-time', 'review-time', 'merging-time', 'release-time'];
+
+    const query = async (interval) => fetchPRsMetrics(
+        api, accountId, 'all', interval, metrics,
+        { repositories: repos, developers: contributors}
+    );
+
+    const currInterval = dateInterval;
+    const prevInterval = {
+        from: dateInterval.from - (dateInterval.to - dateInterval.from),
```
eiso, line 97: This needs to be `dateInterval.from minus one day` to be a true comparison to the previous period.

PR #212, opened 2020-04-09
Add chart descriptions in the info tooltips

warenlg:
https://athenianco.atlassian.net/browse/ENG-460

### DIFF
--- a/src/js/components/insights/stages/merge/mergeDelays.jsx
+++ b/src/js/components/insights/stages/merge/mergeDelays.jsx
@@ -88,7 +88,7 @@ const mergeDelays = {
     factory: (computed) => ({
             meta: {
                 title: 'Merge Delays',
-                description: 'Some description for "Merge Delays"'
+                description: 'Identify in which repositories pull requests take longest from review approval to being merged.'
             },
             content: [
                 {
--- a/src/js/components/insights/stages/review/pullRequestSize.jsx
+++ b/src/js/components/insights/stages/review/pullRequestSize.jsx
@@ -87,7 +87,7 @@ const pullRequestSize = {
     factory: (computed) => ({
             meta: {
                 title: 'Pull Request Size',
-                description: 'Some description for "Pull Request Size"'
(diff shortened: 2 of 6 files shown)


eiso:
I think this comes down to a discussion I thought we could wait with but it will be an important one to have around the products voice: "to whom do we speak and how". Is this for the team or the engineering manager? Is it in parental advisory tone or is it factual professor etc.

rpau:
Yup. Several of the customers calls told us that the prince can be more justified if it is for teams. IMO, this tool is for high performance teams, and manager is another member with a different role.  Is only the manager that should act for merge conflicts or for unreviewed pr's? I do not think so.

There are some potential customers that would like to have "private" dashboards, but this is not our scenario right now, no? (cc @warenlg ) Any chart can be analysed by any member of the team. And, having those private dashboards came to the same question that JIRA, but instead of segmenting by tool, segmenting by team culture.

In order to unblock this PR, and since this is a very simple reversible decision for now, I suggest to **temporary** agree  if we go for **teams** and messages should explain a **purpose**. Do you agree ?

warenlg:
I knew this PR would trigger broader discussions. I quite like the idea of questions btw.
Would you mind letting me merge this PR just to avoid having `Placeholder for the description of chart xyz` ? We can discuss the product wording in general in a dedicated doc or meeting.

warenlg:
Although I quite like the idea of questions as well, I'm wondering if it could be perceived as a bit intrusive.
About descriptions or purposes, tbh I was trying to combine both with small sentences in order to explain the charts (discretely introduce the definition of the metric e.g. PR ratio flow) and suggest how to read them, why they are useful.

warenlg:
I knew this PR would trigger broader discussions.
Most of all my goal with this PR was to get rid off the "placeholder for the description of chart X" 
I did not aim to solve the product wording in general in this quick PR.

warenlg:
Ok, let's discuss this in PRO-19 then.

@dpordomingo ping :)

PR #213, opened 2020-04-10
Change pr number format

vmarkovtsev:
This is just a proposal! https://athenianco.atlassian.net/browse/ENG-534

I changed the PR number format in the list to match GitHub: https://github.com/athenianco/athenian-api/pull/215

Screenshot from 2020-04-10 11-37-58

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "version": "0.1.0",
   "private": true,
   "engines": {
-    "node": ">=12"
+    "node": ">=13"
   },
   "dependencies": {
     "@auth0/auth0-spa-js": "^1.6.3",
--- a/src/js/components/pipeline/PullRequests.jsx
+++ b/src/js/components/pipeline/PullRequests.jsx
@@ -112,10 +112,9 @@ export default ({ stage, data }) => {
                             case 'display':
                                 return `
                                     <div class="table-title">
-                                        <span class="text-secondary">${row.organization}/${row.repo}:</span>
+                                        <span class="text-secondary">${row.organization}/${row.repo}</span>
                                         <a class="text-dark font-weight-bold" href=${github.prLink(row.repository, row.number)} target="_blank">
-                                            <span class="text-secondary">#${row.number}</span>
-                                            ${row.title}
+                                            #${row.number} ${row.title}
                                         </a>
                                     </div>
                                     <div class="table-creators">


warenlg:
I see your point but I'd rather not do it like this as shown in the original ticket https://athenianco.atlassian.net/browse/ENG-534 for these reasons:
* The repo is going to be clickable at some point.
* The PR ID belongs to the PR title, so we want both the PR title and PR ID to follow the same font style (not one grey and the other black). Also to be able to underline both of them on hover

src/js/components/pipeline/PullRequests.jsx
```diff
@@ -112,9 +112,8 @@ export default ({ stage, data }) => {
                             case 'display':
                                 return `
                                     <div class="table-title">
-                                        <span class="text-secondary">${row.organization}/${row.repo}:</span>
+                                        <span class="text-secondary">${row.organization}/${row.repo}#${row.number}</span>
                                         <a class="text-dark font-weight-bold" href=${github.prLink(row.repository, row.number)} target="_blank">
-                                            <span class="text-secondary">#${row.number}</span>
                                             ${row.title}
                                         </a>
```
warenlg, line 118:
```
<span class="text-secondary">${row.organization}/${row.repo}</span>
<a class="text-dark font-weight-bold" href=${github.prLink(row.repository, row.number)} target="_blank">${row.title}#${row.number}</a>
```

vmarkovtsev:
How about removing the double colon then? It does not seem necessary.

src/js/components/pipeline/PullRequests.jsx
warenlg: The PR ID before or after the PR title, I don't mind.

package.json
```diff
@@ -4,7 +4,7 @@
   "version": "0.1.0",
   "private": true,
   "engines": {
-    "node": ">=12"
+    "node": ">=13"
```
dpordomingo, line 7: ehmmmm... what happened?

warenlg:
Yes I prefer @dpordomingo last proposal + removing the double colon after the repo.

warenlg:
Vadim, since you're on holidays and we want to move on with this PR, I asked David to finish the job.
But thanks for opening the PR, this is highly recommended!

package.json
vmarkovtsev: I installed Ubuntu 20.04 :tada: :tada: :tada: It works :+1: :100: :medal_sports:

vmarkovtsev:
Thanks for considering my proposal fellows :smile: Removing the double colon minimizes the visual noise. :+1: 

Something feels *feo* about the numbers being merged with the PR title... For example, if you look at the title of this PR on GitHub, you'll see that the number is colored differently:

image

I don't have any specific proposal here, unfortunately. Adding extra chars? Changing font weight?

PR #220, opened 2020-04-13
Fix review insights

vmarkovtsev:
Fixes:
- ENG-565
- ENG-569


se7entyse7en:
@warenlg this needs you approval given the change in PR size

warenlg:
My bad, I only focused on the `0 file 196 lines` fix. You're right @dpordomingo. May I ask you to fix it, please?
At the end, this number (in the 1st KPI) should be equal to the sum of the files of the yellow bubbles. That's what the users expect I believe. So we do count duplicates btw, since a file under review in 2 different PRs is going to be reviewed 2 times.

PR #223, opened 2020-04-13
Improving the text tooltips

znegrin:
This is related to the issues [#ENG-446](https://athenianco.atlassian.net/browse/ENG-446) and [#ENG-566](https://athenianco.atlassian.net/browse/ENG-566). 

It improves the styles of the text tooltips to make the background white to match the chart tooltips and to make the text larger for a better readability.

image
 
### DIFF
--- a/src/sass/components/_tooltips.scss
+++ b/src/sass/components/_tooltips.scss
@@ -33,3 +33,36 @@
 .user-info {
     font-size: 1.5rem;
 }
+
+// Info tooltips
+
+.tooltip.show {
+    opacity: 1 !important;
+}
+
+.tooltip-inner {
+    max-width: 250px !important;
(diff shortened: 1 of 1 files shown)


znegrin:
> If chart tooltips and these other ones must be similar, why not reusing classes and css definitions?

We were using the existing Bootstrap tooltips, so I just styled them with css as required. I don't know if we can use the Bootstrap tooltips also on the charts.

dpordomingo:
But we can reuse style definitions, can't we?

dpordomingo:
I'm pretty sure the main parts are going to be shared, like shadows, inner colors, borders and such.
Also, reusing definitions do not difficult to adjust some details later; I'd say that even the opposite.
But up to you :(

PR #224, opened 2020-04-13
Fixing the info icon alignment

znegrin:
This is related to the issue [#ENG-567](https://athenianco.atlassian.net/browse/ENG-567).

It fixes the icon misalignment on the main metrics tabs.

image

### DIFF
--- a/src/js/components/pipeline/Thumbnails.jsx
+++ b/src/js/components/pipeline/Thumbnails.jsx
@@ -45,10 +45,12 @@ const Stage = ({ title, text, hint, badge, variation, color, name, metric, activ
         <div className={classnames('card pipeline-thumbnail', active && 'active')} onClick={onClick}>
             <div className="card-body p-3">
                 <div className="card-title d-flex justify-content-between align-items-center mb-3">
-                    <span>
-                        <SmallTitle content={title} />
-                        <Info content={hint} />
-                    </span>
+                    <div className="d-flex justify-content-between align-items-center">
+                        <div className="font-weight-bold text-uppercase text-xs">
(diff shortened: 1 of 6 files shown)


src/js/components/pipeline/Thumbnails.jsx
```diff
@@ -45,10 +45,12 @@ const Stage = ({ title, text, hint, badge, variation, color, name, metric, activ
...
-                    <span>
-                        <SmallTitle content={title} />
-                        <Info content={hint} />
-                    </span>
+                    <div className="d-flex justify-content-between align-items-center">
+                        <div className="font-weight-bold text-uppercase text-xs">
+                            <span className="align-middle">{title}</span>
+                            <Info content={hint} />
+                        </div>
+                    </div>
```
dpordomingo, line 53: `SmallTitle` is already a component to be used everywhere this kind of typography is used, otherwise, we're losing the control of how/where it's used and displayed.
If we need to vertical-align it, we can pass the proper `align-middle` class to that component where needed or modify the component itself to include it always.

znegrin: Then we have to change it, because right now it is not used everywhere, and that's why the icon wasn't aligned in some elements. For example on the `MainMetrics`, and I actually just copied the same code from the main metrics to make it consistent. 
Maybe we can reuse this code for the component instead of the old one.

dpordomingo: If I'm not wrong, this case would be fixed passing `align-middle` className to this instance of `SmallTitle`, won't it?

PR #226, opened 2020-04-13
Fixing sort icon on table columns and order by creation date

znegrin:
This is related to the issue #ENG-568 and it adds a new system to show the sort icons on the table columns to the right of the titles. 

image

image

It also removes the sorting from the first column and adds a new order by default, sorting them by creation date as required by @warenlg.

image

### DIFF
--- a/src/js/components/pipeline/PullRequests.jsx
+++ b/src/js/components/pipeline/PullRequests.jsx
@@ -45,6 +45,7 @@ export default ({ stage, data }) => {
             `,
             searching: true,
             ordering: true,
+            order: [[ 1, 'desc' ]],
             language: {
                 paginate: {
                     next: "<i class='fas fa-angle-right'></i>",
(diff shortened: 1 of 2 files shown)


znegrin:
> This is pretty weird. We're removing the sorting icons from the `th`, and then adding a `span.table-head` and adding the sorting icons there.
> Why not just configuring the original sorting icons to behave as we want?

I tried to position the original icons, but I needed an extra container for the text to do so, as the `th` elements take the full column width so the icons are not positioned where we wanted them to be. That's why I wanted to replace them with the `span` elements. If you have a better idea of how to do this, I would appreciate, as I was dealing with the limitations of the `dataTables` library.

dpordomingo:
I was considering something like this 13da760e1540b80b1dd9ccd749f5c80216873276, wdyt?

znegrin:
> I was considering something like this [13da760](https://github.com/athenianco/athenian-webapp/commit/13da760e1540b80b1dd9ccd749f5c80216873276), wdyt?

Great! Definitely much better! Thank you so much.
I can also merge this to fix and change #227

dpordomingo:
merged

PR #233, opened 2020-04-14
Just ensures that no error is raised in case unable to open installation link

se7entyse7en:
Related to [ENG-338](https://athenianco.atlassian.net/browse/ENG-338). This prevents raising the error and consequently preventing the redirect. In the meantime, in case the link doesn't open, we'll provide the installation link "by hand".


se7entyse7en:
yup, I'll merge this as-is for now and I'll start focusing first on the problem of the first load that requires a refresh.

PR #239, opened 2020-04-15
Show raw ratio of PRs created/reviewed, without gcd

dpordomingo:
required by [[ENG-489]]

[ENG-489]: https://athenianco.atlassian.net/browse/ENG-489


dpordomingo:
By the way, the recursive definition of `gcd` is way cleaner than the deleted one :P

se7entyse7en:
That was the first now I found on SO 😛

PR #242, opened 2020-04-16
Speedup Travis

dpordomingo:
~3mins instead of the current (almost) ~6mins

- Test and build in the same stage, sequentially.
- Avoid building a new docker intermediate image just to build, and do it from dev intermediate image.


se7entyse7en:
Thanks!

PR #243, opened 2020-04-16
Handle missing author in PR size insight chart

se7entyse7en:
## Emulated example screenshot

Screenshot 2020-04-16 at 10 35 22

### DIFF
--- a/src/js/components/insights/stages/review/pullRequestSize.jsx
+++ b/src/js/components/insights/stages/review/pullRequestSize.jsx
@@ -32,7 +32,8 @@ const pullRequestSize = {
                     //   above do not expose it, we can only differentiate between being review-complete, or not.
                     const reviewed = pr.completedStages.includes(PR_STAGE.REVIEW);
 
-                    const author = github.userName(pr.authors[0]);
+                    const authorFullName = pr.authors[0];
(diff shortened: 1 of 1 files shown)


src/js/components/insights/stages/review/pullRequestSize.jsx
```diff
@@ -41,7 +42,9 @@ const pullRequestSize = {
                         number: pr.number,
                         repository: pr.organization + '/' + pr.repo,
                         title: pr.title,
-                        image: github.userImageIndex(data.users)[author],
+                        image: author === 'none' ?
+                            'https://avatars2.githubusercontent.com/u/10137':
```
dpordomingo, line 46: We're already using our own default gh user image for those not having one.([example](https://github.com/athenianco/athenian-webapp/blob/master/src/js/components/layout/Navbar.jsx#L7))

se7entyse7en: ok thanks! @warenlg up to you what you want. Sorry I fixed this really in rush before the demo and we made it by a whisker. Anyway it was due to a 97% fetch of QuickRelease data. But we were lucky to bump into this as it's still a case that can happen and @vmarkovtsev confirmed that this cases exist in go-git for example for a PR whose author deleted the account.

PR #245, opened 2020-04-16
Adding new waiting screen

znegrin:
This is related to the issue [#ENG-612](https://athenianco.atlassian.net/browse/ENG-612). It adds a new temporary waiting screen until we add the future slideshow. Notice that it is supposed to be displayed vertically centered on the full page, so maybe it will need some style adjustments once it is working.

image
Check the video here: https://streamable.com/dbjqga


eiso:
Looks great, can we remove the comma "Please wait"

znegrin:
> Looks great, can we remove the comma "Please wait"
> 
> p.s. Love the effect 😍

I just removed the comma :) I think we could use this same effect to replace the current spinners with skeleton screens in the future. I think it really helps to make it smoother for the users.

eiso:
I think both @warenlg and I agree that skeleton screens are the way to go

dpordomingo:
ready at https://app-staging.athenian.co/prototypes/waiting

PR #254, opened 2020-04-20
Mini charts squash requests

se7entyse7en:
This PR optimizes the calls made to render the mini-charts. Instead of doing 4 separate requests, this PR will now do a single one for all 4 metrics. Side effects:
- the whole row is loaded "at once", but with the skeleton already there
- even if on each filter change the spinner appears, there's still a glitch for the data on the left (the number and the badge variation) as those are data "injected" from outside and not requested by the mini-chart component. I'd rather address this separately as it's part of a bigger change since for now there's not way to know whether those data injected from outside the component are ready or not.

## Gif*

mini-charts-loading-opt

_*done at 5x speed locally_

### DIFF
--- a/src/js/components/DataWidget.jsx
+++ b/src/js/components/DataWidget.jsx
@@ -49,23 +49,29 @@ export default ({id, component, fetcher, plumber, config, propagateSpinner = fal
     }, [id, getData, prevLoadingDataState, loadingDataState]);
 
     const waiting = !dataState;
-    const margin = conf.margin !== undefined ? conf.margin : 5;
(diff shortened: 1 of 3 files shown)


se7entyse7en:
> It is fine to load the 4 metrics of the same row at once.

indeed, actually I think it's even cleaner. 😄 

> Do you have an estimation of the overall speed gain?

It's difficult to say because it depends on the section you're loading since the number of requests is not fixed, but probably if you consider only the mini-charts row, is at least twice faster.

PR #256, opened 2020-04-20
Bump api client

se7entyse7en:
Bumps api client and stop using `granularity`.


se7entyse7en:
@vmarkovtsev after this PR the `granularity` param won't be used anymore for `metrics/prs` endpoint.

vmarkovtsev:
Awesome, thank you Marvin, that was super fast!

I will remove it from the API soon, then we'll see if staging is not broken :+1:

PR #271, opened 2020-04-23
Adding prototype for the Release settings page

znegrin:
This is related to the issue [#ENG-668](https://athenianco.atlassian.net/browse/ENG-668). 

I added the prototype template for the Release settings page.

image

### DIFF
--- a/src/sass/components/_accordions.scss
+++ b/src/sass/components/_accordions.scss
@@ -1,0 +1,29 @@
+.accordion {
+
+    .btn {
+
+        &:focus {
+            box-shadow: none;
+        }
+
+        .accordion-icon {
+            color: $color-secondary;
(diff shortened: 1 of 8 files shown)


src/sass/components/_accordions.scss
```diff
@@ -0,0 +1,29 @@
+.accordion {
+
```
dpordomingo, line 2: These extra lines before every definition are not needed. We use to add them only after a block ends, but not before it starts.

znegrin: Hm, I never leave a space at the beginning of the file so no idea of what happened. Actually in my local copy the file starts on line 1... But will check it twice in the future just in case. Thank you!

PR #281, opened 2020-04-27
Fix PR labeling when no stage has been completed

warenlg:
Fix https://athenianco.atlassian.net/browse/ENG-709
For the same date interval (March 1st to March 31st) before,
before
After,
Screenshot from 2020-04-27 19-42-40

### DIFF
--- a/src/js/services/prHelpers.js
+++ b/src/js/services/prHelpers.js
@@ -270,6 +270,10 @@ export const prLabel = stage => pr => {
 
         return PR_LABELS.RELEASE_PENDING;
     default:
+        if (isStageHappening(PR_STAGE.DONE)) {
+            return PR_LABELS.RELEASE_COMPLETED;
+        }
+
         if (pr.completedStages.length === 0) {
             if (isStageHappening(PR_STAGE.RELEASE)) {
                 return PR_LABELS.RELEASE_PENDING;


src/js/services/prHelpers.js
```diff
@@ -271,7 +271,9 @@ export const prLabel = stage => pr => {
         return PR_LABELS.RELEASE_PENDING;
     default:
         if (pr.completedStages.length === 0) {
-            if (isStageHappening(PR_STAGE.RELEASE)) {
+            if (pr.status === PR_STATUS.MERGED) {
```
se7entyse7en, line 274: Won't this mark as `RELEASED` also PR that are merged, but not release yet? Shouldn't we check for the `done` property and then discriminate whether it is `merged` or `closed`?

warenlg:
I just edited the change according to your comment. However note, that `Done` is not a property but a stage with no end date, so we can not call `stageCompleted` or `stageHappened`. That's why I used `isStageHappening` that is the same in case of `Done` (if it happens, if is completed).

se7entyse7en:
> Also, I'm not checking if the PR is done and then closed because, I believe it is checked earlier https://github.com/athenianco/athenian-webapp/blob/master/src/js/services/prHelpers.js#L232

Ah, indeed 👍

PR #287, opened 2020-04-29
Fix the height of the horizontal bar charts

warenlg:
https://athenianco.atlassian.net/browse/ENG-733

Before,
Screenshot from 2020-04-29 22-13-27
After,
hbar_chart

In particular, with this PR, the chart has the same height as the **Pull Requests Authors** chart in the WIP section, for a given number of users displayed in the y-axis


eiso:
Nice change!

PR #291, opened 2020-04-30
Enable /bearer on prod

vmarkovtsev:
@warenlg, @eiso and I suffer every day: we use the god mode on prod heavily and often change accounts. It is VERY annoying for us to go to the developer tools every time we need an updated Bearer token. The previous assumption that we don't need `/bearer` on prod is completely destroyed, just as I expected.

Please merge this ASAP.

### DIFF
--- a/src/js/Routes.jsx
+++ b/src/js/Routes.jsx
@@ -92,7 +92,7 @@ export default () => {
             <Settings />
           </ContextRoute>
 
-          {devmode && <Route path='/bearer'><Development.Bearer /></Route>}
+          <Route path='/bearer'><Development.Bearer /></Route>
 
           <ContextRoute context={UserContext} path='/waiting'>
             <Waiting />
--- a/src/js/Routes.jsx
+++ b/src/js/Routes.jsx
@@ -39,11 +39,6 @@ export default () => {
         analytics.page();
     }, [location]);
 
-    const devmode = (
-        process.env.NODE_ENV === 'development' ||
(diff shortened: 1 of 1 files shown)


eiso:
Thank you Vadim! This will give me back 5 mins of my day.

src/js/Routes.jsx
```diff
@@ -92,7 +92,7 @@ export default () => {
             <Settings />
           </ContextRoute>
 
-          {devmode && <Route path='/bearer'><Development.Bearer /></Route>}
+          <Route path='/bearer'><Development.Bearer /></Route>
```
se7entyse7en, line 95: Could you remove [this](https://github.com/athenianco/athenian-webapp/pull/291/files#diff-be04a9bedbc0e0c929fb112e61ff0c11R42) as well?

vmarkovtsev: Sorry, the link does not work for me. Do you mean the definition of the `devmode`? I thought it would be useful to keep...

se7entyse7en: > Do you mean the definition of the devmode

yup 👍 

> I thought it would be useful to keep..

It was put there just for this case and it's not a "real" dev-mode it was just actually an arbitrary condition on whether to include or not this path. And also the linter will complain even if we didn't set it up yet in CI.

vmarkovtsev: OK!

vmarkovtsev:
@se7entyse7en Done!

PR #292, opened 2020-05-01
User no account not allowed

se7entyse7en:
Admins onboards will still redirect to waiting page, but it will still be possible to go to home page even without data ready until we don't fully implement progress polling.

## Features

- Better error handling during invitation creation
- New `/signup` route for invitations with update in copies
- Admin account is prioritized in case of multiple accounts
- Authenticaed users without an Athenian account are not allowed
- No more automatic logout in any case
- Disabled showing member invitation link in waiting page
- Styled logout button the same way the login one is

## Videos

- Admin link without error: https://streamable.com/r2v3ra
- Admin link with error: https://streamable.com/vg89ia
- Login without account: https://streamable.com/f4c5iv
- Normal member link without error: https://streamable.com/y0077b
- Normal member link with error: https://streamable.com/2jjfx9
- Signup without link: https://streamable.com/jo3jko

### DIFF
--- a/src/js/Routes.jsx
+++ b/src/js/Routes.jsx
@@ -17,6 +17,7 @@ import Pipeline from 'js/pages/Pipeline';
 import Stage from 'js/pages/pipeline/Stage';
 import Overview from 'js/pages/pipeline/Overview';
 import Login from 'js/pages/auth/Login';
+import SignUp from 'js/pages/auth/SignUp';
 import Logout from 'js/pages/auth/Logout';
 import Development from 'js/components/development';
 import Callback from 'js/pages/auth/Callback';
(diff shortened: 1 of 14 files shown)


se7entyse7en:
> In the 1st video, where is pointing the Athenian logo both on the waiting page? Can the user access the overview page through there, before installing the app or fetching the entire org?

Totally. As I said in the description, until we don't start polling the `progress` endpoint there's no way to "block" admins for jumping into the overview section.

warenlg:
Ok I see, I'll follow up on this ticket then https://athenianco.atlassian.net/browse/ENG-301
Otherwise, this sounds really good to me. Thank you Marvin.

se7entyse7en:
> One near term task I would add is logging these errors to Sentry

Totally agree, we've been dealing with that many things that the PR for sentry is still there... 😓

PR #293, opened 2020-05-01
Fix some random errors that broke the page

dpordomingo:
- b37ba55 Handle no data situations, e.g. ancient dates
- a71e18d Use format.number.fixed(n,d) instead of n.toFixed(d)


dpordomingo:
@vmarkovtsev you might be interested in 553d93d since you introduced some `toFixed`.

### TL;DR
`toFixed` is a method of `Number`, so it won't be available on `null` or `Strings`
use `format::number.fixed(n,d)` as in 553d93d

vmarkovtsev:
Thanks for educating David :+1: 

Why are there possibilities to work with a `null` instead of a number? When the API returns nothing?

PR #297, opened 2020-05-04
Settings

dpordomingo:
[[ENG-696]] Add the release settings page into the current settings of the webapp

Add Teams and Releases settings. Only FE, no connection with the API.

image

[ENG-696]: https://athenianco.atlassian.net/browse/ENG-696


dpordomingo:
@znegrin I moved the code from the prototypes to functional code in `/settings`.
As you can see, the layout behaves weird when the tag strategy is chosen, or when the default branch is not called `master`.

Also, I had to replace the Bootstrap Dropdown because it's jQuery based. We're already using `react-select`. I think you could be also interested in styling it.

I'm merging this to be accessible at staging.

PR #299, opened 2020-05-04
feat(ENG-687): Add developer names to filter

navassouza:
Closes: https://athenianco.atlassian.net/browse/ENG-687

Only show the extra name if login and name are differente. example attached: dependabot-preview does not have an extra name since it'll be dependabot-preview (dependabot-preview)

Screen Shot 2020-05-04 at 17 58 07


se7entyse7en:
As a side note, I'll create a ticket for configuring `prettier` and `eslint`, but low priority.

PR #300, opened 2020-05-04
feat(ENG-747): Add tickformat to Review Activity

navassouza:
Closes: https://athenianco.atlassian.net/browse/ENG-747

Add prop tickFormat

Steps:

1) http://localhost:3000/stage/review
2) Scroll to insights
3) Horizontal bar chart now has '%' in its tickformat

### DIFF
--- a/src/js/components/insights/stages/review/reviewActivity.jsx
+++ b/src/js/components/insights/stages/review/reviewActivity.jsx
@@ -212,6 +212,7 @@ const reviewActivity = {
                         params: {
                             title: 'Most Active Reviewers',
                             data: computed.secondBox.chartData,
+                            tickFormat: '%',
                             extra: {
                                 yAxis: {
                                     imageMapping: computed.secondBox.avatarMapping,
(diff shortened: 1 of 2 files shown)


eiso:
Really like your way of naming PRs. Haven’t seen feat(jiraID) before, it’s great!

navassouza:
@eiso,
Mozio had some strict rules to create PR, it’s just a habit now. I’m more than glad to share a lot of ideas regards this

Cheers

src/js/components/insights/stages/review/reviewActivity.jsx
```diff
@@ -212,6 +212,7 @@ const reviewActivity = {
                         params: {
                             title: 'Most Active Reviewers',
                             data: computed.secondBox.chartData,
+                            tickFormat: '%',
```
se7entyse7en, line 215: If we wanna call this `tickFormat` I'd make this a function, so the same you pass to the `XAxis` component: `(v) => ${v}${tickFormat}`. You'll then be able to just pass it without checking `tickFormat?`.

Otherwise, I'd suggest something like `tickTextSuffix: '%'`.

navassouza: agreed, a func would be more suitable.

navassouza: Done.
1) added a default value for tickFormat in case prop is undefined
2) checked other Horizontal bar chart making sure it would break anything

PR #302, opened 2020-05-05
Fix filter values passed to api

se7entyse7en:
Fixes the values passed to api for developers filter introduced by #299.


se7entyse7en:
I'll merge once CI done to unblock staging.

PR #312, opened 2020-05-06
fix(ENG-749): Replace Reviews count by Prs Reviewed

navassouza:
Closes: https://athenianco.atlassian.net/browse/ENG-749

Replace `reviews` metric by `prs-reviewed`

### DIFF
--- a/src/js/components/insights/stages/review/reviewActivity.jsx
+++ b/src/js/components/insights/stages/review/reviewActivity.jsx
@@ -13,9 +13,8 @@ const reviewActivity = {
     fetcher: async (api, context) => {
         const fetchFirstBox = async () => {
             const metrics = [
-                'reviews',
-                'prs-created',
                 'prs-reviewed',
+                'prs-created',
             ];
 
             return fetchDevsMetrics(
@@ -27,7 +26,7 @@ const reviewActivity = {
 
         const fetchSecondBox = async () => {
             const metrics = [
-                'reviews',
+                'prs-reviewed',
                 'pr-comments',
             ];
 
@@ -53,7 +52,7 @@ const reviewActivity = {
               .map(v => v.values[0][1])
               .sum();
         const totalReviewedPRs = _(data.firstBox.calculated)
-            .map(v => v.values[0][2])
+            .map(v => v.values[0][1])
             .sum();
         const totalReviewers = _(data.secondBox.calculated)
             .filter(v => v.values[0][0] > 0)
@@ -156,7 +155,6 @@ const reviewActivity = {
     factory: (computed) => {
         const createdPRs = computed.firstBox.KPIsData.createdPRs;
         const reviewedPRs = computed.firstBox.KPIsData.reviewedPRs;
-
         return {
             meta: {
                 title: 'Review Activity',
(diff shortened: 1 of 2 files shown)


src/js/components/insights/stages/review/reviewActivity.jsx
```diff
@@ -27,7 +26,7 @@ const reviewActivity = {
 
         const fetchSecondBox = async () => {
             const metrics = [
-                'reviews',
+                'prs-reviewed',
```
se7entyse7en, line 29: Can you update the variable names accordingly to make the code consistent? As an example in L:63.

src/js/components/insights/stages/review/reviewActivity.jsx
```diff
@@ -13,9 +13,8 @@ const reviewActivity = {
     fetcher: async (api, context) => {
```
se7entyse7en, line 13: This can be refactored using a single call to `fetchDevsMetrics`, but I don't mind doing it in a separate PR.

warenlg:
I actually raised the same 2 concerns to Ricardo during a sync call. You're right @se7entyse7en let's address them now.

src/js/components/insights/stages/review/reviewActivity.jsx
navassouza: done

src/js/components/insights/stages/review/reviewActivity.jsx
navassouza: not sure about the var names, but refactored the code to use less 'magic numbers' and make it clearer

src/js/components/insights/stages/review/reviewActivity.jsx
```diff
@@ -1,3 +1,4 @@
+// @flow
```
se7entyse7en, line 1: Since we're not using `flow` can you remove this directive? I don't have anything against it, but it's part of a larger discussion. We have a ticket to discuss `eslint` and `prettier` as well, we could include `flow` as well. Related to this there were some light discussions around `prop-types` \cc @dpordomingo.

Also for any of this things we'll be useful to include the corresponding checks in CI.

I also don't see any flow annotations here, how are you using it?

navassouza: oh, I want to start using it in this project, but since it's for a future discussion, I can remove for now. 
I added just for the sake of catching any error I wasn't being able to see.

It works even without the annotations, so I can add them incrementally.
For clarification, see example #2: https://flow.org/en/docs/getting-started/

Since refactoring may cause breaks, I just wanted to have a sanity check in the meantime.

navassouza: prop-types are discouraged by react team, they suggest either flow or typescript, I'm fine working with any, but typescript would have a steep learning curve for the team. So flow is the way to go

navassouza: `Static type checkers like Flow and TypeScript identify certain types of problems before you even run your code. They can also improve developer workflow by adding features like auto-completion. For this reason, we recommend using Flow or TypeScript instead of PropTypes for larger code bases.`

se7entyse7en: I totally agree with you wrt to `flow` vs `ts`. 👍 Let's just discuss it in the ticket instead of introducing it in a random PR.

wrt to `prop-types` also that works only for React components while `flow` is more generic, so I agree that it's the way to go since it doesn't require blocking development but can be added incrementally.

dpordomingo: `flow` ftw

PR #314, opened 2020-05-07
feat(ENG-778): Add filter by status to PRs table

navassouza:
Closes: https://athenianco.atlassian.net/browse/ENG-778

### DIFF
--- a/src/js/components/pipeline/PullRequests.jsx
+++ b/src/js/components/pipeline/PullRequests.jsx
@@ -12,257 +12,258 @@ import _ from 'lodash';
 import { StatusIndicator, READY } from 'js/components/ui/Spinner';
 
 const userImage = users => user => {
-    if (users[user] && users[user].avatar) {
-        return `<img
-            src="${users[user].avatar}"
-            title="${github.userName(user)}"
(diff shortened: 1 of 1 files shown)


src/js/components/pipeline/PullRequests.jsx
```diff
@@ -1,268 +1,303 @@
...
+  const filterOptions = [
+    { label: 'All', value: null },
+    ...Object.entries(PR_LABELS).map(([key, value]) => ({ label: value, value }))
+  ]
+
+  return (
+    <>
+      <StatusIndicator status={status} textOnly={false} />
+      <div className="table-responsive mb-4">
+        <div className="d-flex" style={{ marginBottom: '-34px', justifyContent: 'flex-end' }}>
```
navassouza, line 64: I'm not a big fan os leaving style like this, but in this case is a bit exceptional

navassouza:
@warenlg had some design constraints about size, so we opted not to use multiselect, just a plain single select

navassouza:
@se7entyse7en, after adding the status filter got search, paging and ordering tested. we should be good

PR #315, opened 2020-05-08
Better sentry errors

se7entyse7en:
Closes [ENG-784](https://athenianco.atlassian.net/browse/ENG-784).

Adds a wrapper to use on all api requests in order to be able to track errors on Sentry due to non-200 responses.

### DIFF
--- a/src/js/services/api/index.js
+++ b/src/js/services/api/index.js
@@ -8,6 +8,7 @@ import {
   CalculatedPullRequestMetrics,
   ReleaseMatchRequest,
 } from 'js/services/api/openapi-client';
+import * as Sentry from '@sentry/browser';
 import ForSet from 'js/services/api/openapi-client/model/ForSet';
 import PullRequestMetricID from 'js/services/api/openapi-client/model/PullRequestMetricID';
 import DeveloperMetricID from 'js/services/api/openapi-client/model/DeveloperMetricID';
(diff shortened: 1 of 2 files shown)


se7entyse7en:
@dpordomingo @navassouza if you happen to add some api calls please remember to wrap them with this Sentry capturer.

src/js/services/api/index.js
```diff
@@ -8,15 +8,55 @@ import {
...
+    }
+}
+
+export const reportToSentry = (error, context = {}) => Sentry.withScope(scope => {
+    scope.setTags(context.tags);
+    scope.setExtras(context.extra);
+    Sentry.captureException(error);
+});
+
+const withSentryCapture = (p, message) => p.catch(e => {
```
dpordomingo, line 41: Catching the error causes that the caller won't realize about a failed request.
That issue was a blocker for #317 so I patched it by https://github.com/athenianco/athenian-webapp/pull/317/commits/3107ceacbf2669c1b317e47341d80e35b1c50e00
I guess that in the rest of the current methods this is not an issue yet, but something to consider later when https://athenianco.atlassian.net/browse/ENG-506

se7entyse7en: It makes a lot of sense to me 👍, thanks!

PR #316, opened 2020-05-08
fix(): Filter cropping/Filter options by context

navassouza:
Fixes Filter of PRs

### DIFF
--- a/src/js/components/pipeline/PullRequests.jsx
+++ b/src/js/components/pipeline/PullRequests.jsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useState, useMemo } from 'react';
 import $ from 'jquery';
 import 'datatables.net';
 import 'datatables.net-bs4';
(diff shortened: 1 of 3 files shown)


src/js/components/pipeline/PullRequests.jsx
```diff
@@ -30,12 +30,29 @@ const tableContainerSelector = `#${tableContainerId}`;
...
+    return [
+      { label: 'All' },
+      ...labels.map(value => ({ value, label: value }))
+    ]
+  }, [data.prs, stage])
+
+  const isReady = status => status === READY
+
+  // reset filter when changing stage
+  useEffect(() => setSelectValue(null), [stage])
```
se7entyse7en, line 47: Is this working as expected? I guess you meant:

```suggestion
  useEffect(() => {
      setSelectValue(null)
  }, [stage])
```
?

se7entyse7en: I think the currently you're returning it to be run as cleanup function, but you're not actually returning a function so React will complain

navassouza: true, for a moment forgot it's an implicit return, fixed

PR #317, opened 2020-05-11
Read API errors

dpordomingo:
Errors returned by the API, contains their details under error.body
Errors caused while parsing the response (e.g. JSON decode errors), are under err.error
Plain errors are also stringed with its original cause, if it exists
Strings and numbers are left untouched

### DIFF
--- a/src/js/services/api/index.js
+++ b/src/js/services/api/index.js
@@ -38,7 +38,7 @@ export const reportToSentry = (error, context = {}) => Sentry.withScope(scope =>
     Sentry.captureException(error);
 });
 
-const withSentryCapture = (p, message) => p.catch(e => {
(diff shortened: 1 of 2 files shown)


vmarkovtsev:
Note: the server can return an empty response - close the connection without even setting and HTTP status code. One example that I know is debugging it locally and exiting. It may happen on a SIGSEGV crash in a native extension, too. I think it should be considered as 500.

dpordomingo:
Thanks @vmarkovtsev for your example. I could not reproduce it, but I guess that it would be a similar example if I kill the API before calling it, so I get this error
image

vmarkovtsev:
Yes, it should be the same.

PR #319, opened 2020-05-12
Enable 'Settings' in user menu. Hide 'Teams' menu

dpordomingo:
This PR will enable the link to `Settings` in the user menu, and hide the `Teams` link in the Settings aside in Production.
https://app.athenian.co/settings/team route will still be accessible by direct url.

image

image

### DIFF
--- a/src/js/pages/Settings.jsx
+++ b/src/js/pages/Settings.jsx
@@ -25,6 +25,8 @@ export const getOrg = user => {
 export default ({ children }) => {
   const { user } = useUserContext();
 
+  const isDev = window.ENV.environment === 'development' || process.env.NODE_ENV === 'development';
+
   if (!user) return <Page />;
 
   return (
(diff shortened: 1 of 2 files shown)


dpordomingo:
Let me know @warenlg if we should deploy this into Staging, Prod or both.

warenlg:
Staging first. I'll check that everything works correctly there and if all good -> prod

src/js/pages/Settings.jsx
```diff
@@ -25,6 +25,8 @@ export const getOrg = user => {
 export default ({ children }) => {
   const { user } = useUserContext();
 
+  const isDev = window.ENV.environment === 'development' || process.env.NODE_ENV === 'development';
```
se7entyse7en, line 28: Not a blocker, but including also `window.ENV.environment === 'staging'` could be handy.

dpordomingo: good catch! added by https://github.com/athenianco/athenian-webapp/pull/320

PR #321, opened 2020-05-13
refactor(ENG-291): Re-style filter dropdowns

navassouza:
Closes: https://athenianco.atlassian.net/browse/ENG-291
https://athenianco.atlassian.net/browse/ENG-824
Screen Shot 2020-05-12 at 22 11 43


src/js/components/ui/filters/MultiSelect/CustomComponents.jsx
```diff
@@ -0,0 +1,181 @@
...
+    }
+  }, [])
+
+  const childrenProps = {
+    menuIsOpen: isMenuOpen,
+    components: {
+      Option,
+      Placeholder: placeholder(label),
+      Menu: menu({ setMenuOpen, onApply })
+    },
```
se7entyse7en, line 44: Cool! I wasn't aware that `react-select` lets you chang the built-ins 👍

src/js/components/ui/filters/MultiSelect/CustomComponents.jsx
```diff
@@ -0,0 +1,181 @@
...
+  )
+}
+
+/**
+ * Menu
+ */
+export const menu = ({ setMenuOpen, onApply }) => props => {
+  const {
+    getStyles,
+    innerProps: { ref, ...restInnerProps },
```
se7entyse7en, line 133: Is this `ref` part of the thing passed internally by `react-select`?

src/js/components/ui/filters/MultiSelect/CustomComponents.jsx
```diff
@@ -0,0 +1,181 @@
...
+  } = props
+
+  const style = {
+    ...getStyles('menu', props),
+    boxShadow: 'none',
+    border: '1px solid #ccc',
+    borderTopWidth: 0,
+    cursor: 'pointer'
+  }
+  const allValues = getValue()
```
se7entyse7en, line 148:
```suggestion
  const selectedValues = getValue()
```
wdyt?

src/js/components/ui/filters/MultiSelect/CustomComponents.jsx
```diff
@@ -0,0 +1,181 @@
...
+  const allSelected = allValues.length === options.length
+
+  return (
```
se7entyse7en, line 151: @navassouza correct me if I'm wrong, but so currently a change is applied only on `Apply` and in case the user clicks outside the dropdown it is just closed, but not applied, am I correct?

src/js/components/ui/filters/MultiSelect/CustomComponents.jsx
```diff
@@ -0,0 +1,181 @@
...
+        <span>
+          <span className="filter-dropdown-all">All</span>
+          <span className="filter-dropdown-pill">
+            {options.length}
+          </span>
+        </span>
+      </div>
+      {children}
+      <FilterFooter
+        onCancel={() => setMenuOpen(false)}
```
se7entyse7en, line 172: Does it make sense to restore previous values? It's not done here AFAIS. \cc @warenlg

src/js/pages/pipeline/Filters.jsx
```diff
@@ -1,224 +1,254 @@
...
+  const [filteredReposState, setFilteredReposState] = useState([])
+
+  const [contribsReadyState, setContribsReadyState] = useState(false)
+  const [allContribsState, setAllContribsState] = useState([])
+  const [filteredContribsState, setFilteredContribsState] = useState([])
+
+  const [filteredDateIntervalState, setFilteredDateIntervalState] = useState(defaultDateInterval)
+  
+  const [reposSelectState, setReposSelectState] = useState([])
+  const [usersSelectState, setUsersSelectState] = useState([])
```
se7entyse7en, line 47: At some point it would be better to refactor this using `useReducer` it will simplify state changes especially given that currently a change in a filter affects a change in the others. By using `useReducer` will be able to reduce the number of state changes.

src/js/pages/pipeline/Filters.jsx
```diff
@@ -1,224 +1,254 @@
-import React, { useState } from 'react';
+import React, { useState, useEffect } from 'react'
```
se7entyse7en, line 1: I didn't check this extensively since more or less seems the same.

se7entyse7en:
@navassouza please next time wait for at least approval from me or @dpordomingo for code review. If you could answer the CR comments even post-merge that would be appreciated.

Anyway good work overall 👍

navassouza:
> Great job on this @navassouza it renders really well, let's see how it behaves on staging.
> 
> I have some questions and remarks though:
> 
> * Why the filter box does not have the same width as the breakdown which seems a bit larger?
> 
> * What is the hover effect when you go through the devs in the list?
> 
> * Does the search applies to both full names and Github handles?
> 
> * Does the user have to click inside the checkbox or clicking anywhere in the row tick the checkbox of the corresponding dev?

1) It’s in the design proposal (attached) image

2) no hover effect, but simple to add

3) yes, I fixed this use case, we can search by voth nane and gh handle

4) click anywhere

src/js/components/ui/filters/MultiSelect/CustomComponents.jsx
```diff
@@ -0,0 +1,181 @@
...
+    </div>
+  )
+}
+
+/**
+ * Placeholder,
+ * @param {string} label
+ * @return {function} 
+ */
+export const placeholder = label => props => {
```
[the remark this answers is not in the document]
navassouza: Ops, not really

src/js/components/ui/filters/MultiSelect/CustomComponents.jsx
navassouza: Yes

src/js/components/ui/filters/MultiSelect/CustomComponents.jsx
navassouza: Change is applied only after click apply, click outside just closes the dropdown

src/js/components/ui/filters/MultiSelect/CustomComponents.jsx
navassouza: Nvm, you’ll see some random changes like this when I’m reading code and trying to understand it

src/js/context/Filters.jsx
```diff
@@ -1,13 +1,13 @@
...
+export const useFiltersContext = () => useContext(Context);
 
-export default ({ ready, repositories, contributors, dateInterval, children }) => {
-    return (
-        <FiltersContext.Provider value={{ ready, repositories, contributors, dateInterval }}>
-            {children}
-        </FiltersContext.Provider >
-    );
-};
+export default function FiltersContext({ ready, repositories, contributors, dateInterval, children }) {
```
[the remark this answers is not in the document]
navassouza: Yes, when checking React debugger anonymous functions show as ‘Anonymous’ and I can’t know for sure what component is that, so we need a named one or add .displayName manually everywhere. Either arrow/named function works, no preference for one over another here

src/js/pages/pipeline/Filters.jsx
navassouza: Yes, you’re right. Most changes here are indentation changes I do while reading

src/js/pages/pipeline/Filters.jsx
navassouza: Agreed

src/js/components/ui/filters/MultiSelect/CustomComponents.jsx
warenlg: Ah when onCancel ok. Yes absolutely, the cancel button should close the filter and reset the variables https://athenianco.atlassian.net/browse/ENG-850

PR #324, opened 2020-05-13
Handle empty badges

dpordomingo:
addresses [[ENG-477]] Remove the diff badges when there is no data to compare  
addresses [[ENG-481]] NaN in the "Proportion of the cycle time"
addresses [[ENG-779]] Fix the NAN and style the tooltip of the Lead Time distribution chart

`Badge` component is now responsible for rendering the badge if it received a number (or also string if it is not in `trend` mode). If it received a `NaN` or a `+-Infinite` it will send it as an error to Sentry because it will mean that a division by zero, or a wrong division happened.

You can see below the before, and after (in not Prod env; if it would be rendered in Prod, the error highlight won't be visible)
image

Also, **when it is not in Production**, such `NaN` or a `+-Infinite` errors will be highlighted as in the following screenshot, letting the developer or Product know that something was not properly calculated.
image

[ENG-477]: https://athenianco.atlassian.net/browse/ENG-477
[ENG-481]: https://athenianco.atlassian.net/browse/ENG-481

[ENG-779]: https://athenianco.atlassian.net/browse/ENG-779

### DIFF
--- a/src/js/components/development/index.jsx
+++ b/src/js/components/development/index.jsx
@@ -59,5 +59,7 @@ export default {
       (async () => setToken(await getTokenSilently()))();
 
       return <Simple><pre className="bearer-token">{token}</pre></Simple>;
-    }
+    },
+
+    errorBoxClass: isNotProd && 'development-error-box',
 };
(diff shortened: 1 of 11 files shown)


se7entyse7en:
Cool trick highlighting errors 👍, as a side note to @warenlg would it make sense to add the transparency to the bubbles? In the example, in the screenshot, it's difficult to understand how many bubbles are there.

PR #331, opened 2020-05-15
fix(ENG-776): Fix flashing no-data placeholder

navassouza:
Closes: https://athenianco.atlassian.net/browse/ENG-776

### DIFF
--- a/src/js/pages/pipeline/Body.jsx
+++ b/src/js/pages/pipeline/Body.jsx
@@ -1,187 +1,182 @@
-import React, {useEffect} from 'react';
-import { useParams } from "react-router-dom";
-import _ from "lodash";
-import moment from 'moment';
+import React, {useEffect} from 'react'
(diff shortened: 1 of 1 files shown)


src/js/pages/pipeline/Body.jsx
```diff
@@ -1,187 +1,182 @@
...
+                  
+      setGlobalData('prs-metrics.variations', fetchValues())
+    }
+                
+    getAndSetPRs()
+    fetchGlobalPRMetrics()
+    fetchGlobalPRMetricsVariations()
+  }, [apiReady, api, apiContext.account, apiContext.interval, apiContext.repositories, apiContext.contributors, setGlobalData])
+
+  if (!apiReady) {
```
navassouza, line 157: The fix is from this line and below, I usually re-indent lines while reading >.<

PR #332, opened 2020-05-15
[HOTFIX] Fix missing .value() call to loadash wrapper

se7entyse7en:


se7entyse7en:
This is a hotfix to cope with the semantic change in the `properties` returned by the api. `prHelpers` has to be re-checked to properly apply the new "no time machine" semantic.

PR #339, opened 2020-05-19
feat(ENG-822): Add the drop-down listing users when creating a team i…

navassouza:
…n the settings


src/js/services/api/openapi.1.0.41.yaml
```diff
@@ -1425,7 +1638,10 @@ components:
           description: Name of the team.
           type: string
         members:
+<<<<<<< HEAD:src/js/services/api/openapi.1.0.43.yaml
```
dpordomingo, line 1641: Ouch.

src/js/pages/settings/Team.jsx
```diff
@@ -1,54 +1,164 @@
...
+      name,
+      members: members.map(m => m.login)
+    }
+    try {
+      const { id } = await createTeam({ token, body })
+      setTeams([
+        { id, members, name },
+        ...teams,
+      ])
+    } catch (err) {}
```
dpordomingo, line 69: We should let the user know that something went wrong.
example https://github.com/athenianco/athenian-webapp/blob/master/src/js/pages/settings/Releases.jsx#L123

src/js/services/api/index.js
```diff
@@ -164,9 +164,36 @@ export const getContributors = (token, userAccount, from, to, repos) => {
...
         {from: new Date(from), to: new Date(to)},
         {repositories:repos}
-    ).then(contribs => contribs.map(({ login, name, avatar }) => ({ login, name, avatar })));
+    )
+    .then(contribs => contribs.map(({ login, name, avatar }) => ({ login, name, avatar })));
 };
 
+export const getDevelopers = (token, userAccount) => {
+  const api = buildApi(token)
+  return api.getContributors(userAccount)
```
dpordomingo, line 173: we must use sentry wrapper for error handling.
example https://github.com/athenianco/athenian-webapp/blob/master/src/js/services/api/index.js#L377

src/js/services/api/openapi.1.0.41.yaml
dpordomingo: I just realized that `src/js/services/api/openapi.1.0.41.yaml` should be fully deleted as done when the API is upgraded using `make generate` (which replaces the old specs with the new ones)

src/js/pages/settings/Team.jsx
```diff
@@ -1,54 +1,164 @@
...
+}
+
+const getOptionValueUsers = val => `${val.name} ${val.login}`
+
+const plusIcon = String.fromCharCode(43)
+
+const closeDropdown = () => document.body.click()
+
+export default function Teams() {
+  const { getTokenSilently } = useAuth0()
```
dpordomingo, line 48: Our components must retrieve the api service from the context as in this example.
https://github.com/athenianco/athenian-webapp/blob/master/src/js/pages/settings/Releases.jsx#L31

This way we also ensure that user token is sent to the API avoiding the bug described in this slack message.
https://athenian.slack.com/archives/CQUFEUP5H/p1589881146080900

PR #342, opened 2020-05-19
New alert styling and copy

znegrin:
This is based on the issue [#ENG-838](https://athenianco.atlassian.net/browse/ENG-838), and it adds new styles and behavior for the `Toastify` alerts and a new copy for the success cases to simplify the messaging. 

It is still missing the new copy for the errors as it had several conditions that I wasn't sure about.

image

image

### DIFF
--- a/src/js/App.jsx
+++ b/src/js/App.jsx
@@ -1,7 +1,7 @@
 /* global process */
 import React from 'react';
 import ReactDOM from 'react-dom';
-import { ToastContainer, toast } from 'react-toastify';
+import { ToastContainer, toast, Slide } from 'react-toastify';
 
 import {
     Router,
@@ -63,7 +63,16 @@ ReactDOM.render(<>
       </Router>
     </React.StrictMode>
   </Auth0Provider>
-  <ToastContainer />
+  <ToastContainer
+      position="top-center"
+      hideProgressBar
+      closeOnClick
+      pauseOnFocusLoss={true}
+      draggable={false}
+      pauseOnHover={true}
+      transition={Slide}
+      closeButton={false}
+  />
   {process.env.NODE_ENV === 'development' && <Development.Breakpoints />}
 </>,
   document.getElementById('root')
--- a/src/js/App.jsx
+++ b/src/js/App.jsx
@@ -66,12 +66,12 @@ ReactDOM.render(<>
   <ToastContainer
       position="top-center"
       hideProgressBar
+      closeButton={false}
       closeOnClick
-      pauseOnFocusLoss={true}
       draggable={false}
+      pauseOnFocusLoss={true}
       pauseOnHover={true}
       transition={Slide}
-      closeButton={false}
   />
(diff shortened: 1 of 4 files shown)


src/js/App.jsx
```diff
@@ -63,7 +63,17 @@ ReactDOM.render(<>
       </Router>
     </React.StrictMode>
   </Auth0Provider>
-  <ToastContainer />
+  <ToastContainer
+      position="top-center"
+      autoClose={5000}
+      hideProgressBar
+      closeOnClick
+      pauseOnFocusLoss={false}
```
dpordomingo, line 71: are we really sure about this? It's really annoying when a message disappears without being able to read or process or take a screenshot. 
Same about `pauseOnHover={false}`.
I strongly recommend keeping both.

src/js/App.jsx
```diff
@@ -63,7 +63,17 @@ ReactDOM.render(<>
       </Router>
     </React.StrictMode>
   </Auth0Provider>
-  <ToastContainer />
+  <ToastContainer
+      position="top-center"
+      autoClose={5000}
```
dpordomingo, line 68: Imo 5s for a success message can be too much, but let's see.

dpordomingo, line 68: Imo 5s for a success message is too much.

src/js/App.jsx
dpordomingo, line 71: are we really sure about this? It's really annoying when an error message disappears without being able to read or process or take a screenshot. 
Same about `pauseOnHover={false}`.
I strongly recommend keeping both.

src/js/App.jsx
warenlg: Yes, let's make it shorter. I'm ok with 1s for the success message

src/js/App.jsx
warenlg: We can keep both, yes.

warenlg:
I'm against having specific messages with proper values because it yields to 4 lines error or success messages. Let's simplify the error messages, please. If we reach a ForbiddenError, then let's display "Permission denied" for instance.

dpordomingo:
@znegrin If you don't feel comfortable adding the error handling that @warenlg requested by https://github.com/athenianco/athenian-webapp/pull/342#issuecomment-631344722, assign me the issue once this is merged, and once you're ready with your part. :)

znegrin:
@dpordomingo I agree with your suggestions. I just added specific values to the messages on `logger.jsx` but it seems that it is not working. Can you please review it in case I am missing something?

For the specific messages for the errors, I didn't change them as I wasn't sure about the meaning of each case, so if you can please help me with that too I would appreciate it. Thank you!

dpordomingo:
I started working on this.
Will rebase and add error handling :D

dpordomingo:
Error messages are now shorter.

I reproduce in the following screenshots the responses that the API can return (but the `422`, which I could not reproduce)

image

Imo with this error messages the user will be able to know what happened, and will be able to let us know as well.

PR #346, opened 2020-05-20
.env file to fix eslint errors from open-api generator

navassouza:
We currently have `eslint` running by default cause of CRA, but we can't use `.eslintignore` on open-api generator files, which causes a list of error logs all the time and is annoying to deal with during development.

I see in `package.json` we tried ignoring the whole folder, but it's not working because we need to enable it first.

// packache.json config to ignore the folder
`"eslintConfig": {
    "extends": "react-app",
    "ignorePatterns": [
      "openapi-client"
    ]
  },`

We can enable it in 2 ways:

Edit `"start": "react-scripts start"` to `"start": "EXTEND_ESLINT=true react-scripts start"` **OR** add the `.env` file

https://github.com/facebook/create-react-app/commit/6f5221c2d7df0b7a097cfdd461967422c3013e12

This PR has a .env file and commented out .ignore of .env file


navassouza:
@dpordomingo, I get this while coding. The problem probably is I ran `npm install` and all the setup manually instead of using the Makefile
Can you confirm?
Screen Shot 2020-05-21 at 10 06 49

dpordomingo:
Yes, i think that `make serve`, as said in our [docs](https://github.com/athenianco/athenian-webapp/blob/master/docs/CONTRIBUTING.md#available-scripts), is what you need.

PR #348, opened 2020-05-21
No god tracking

se7entyse7en:
Closes [ENG-629](https://athenianco.atlassian.net/browse/ENG-629).

### DIFF
--- a/src/js/analytics.js
+++ b/src/js/analytics.js
@@ -10,9 +10,9 @@ const getAnalytics = () => {
         analytics.page = window.analytics.page;
         analytics.track = window.analytics.track;
     } else {
-        analytics.identify = (...args) => {console.log(`dummy analytics: identify [${args}]`);};
-        analytics.page = (...args) => {console.log(`dummy analytics: page [${args}]`);};
-        analytics.track = (...args) => {console.log(`dummy analytics: track [${args}]`);};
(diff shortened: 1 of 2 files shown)


eiso:
Great! Happy to start having accurate usage metrics

eiso:
One question, I see you're using identify but not disabling segment. Does this mean events will still be sent to mixpanel but just unidentified? And same for intercom? Etc 

This should be fine but want to test that intercom doesn't leave a cookie.

se7entyse7en:
> One question, I see you're using identify but not disabling segment. Does this mean events will still be sent to mixpanel but just unidentified? And same for intercom? Etc

Yup, in this PR I just avoid identifying the user instead of disabling segment for different reasons:
- segment is loaded on page load,
- we track the page load before the user authenticates (so even in god mode).

Alternatives:
- load segment after authentication and after check if god mode or not but it's not how their integration doc suggests, and this means we'll lose tracking some page loads which I don't personally like,
- load segment on load as now, but don't track anything until not authenticated which has the same problem of losing some page loads and potentially also information about referrers.

> This should be fine but want to test that intercom doesn't leave a cookie.

Let's test this in staging.

se7entyse7en:
@eiso added you as reviewer as well.

eiso:
Catch 22: approval needed to go into staging, can only approve once tested in staging ;)

PR #349, opened 2020-05-21
Use TimeSeries chart type instead of dupe FilledAreaChart

dpordomingo:
partially addresses [[ENG-537]] Ignore points for days of the charts when we don't have data

This PR also removes `FilledAreaChart` and use `TimeSeries` chart instead, which was already available and properly handling null-values.

After:
image

Before:
image

[ENG-537]: https://athenianco.atlassian.net/browse/ENG-537


dpordomingo:
@warenlg since `null` values are not painted, it may happen that the first day is not in the line, like in this example.
image

If we really need to change this, I'd do it in another PR

warenlg:
Yes this is what I mentioned in the ticket.

> Also, there might be some corner cases to handle properly, like having data for the days 1, 2, 3 and 4 but not for the rest of the week. In that case, I’d suggest to just cut the curve, and draw the line only from day 1 to day 4

Let's try it like this for now and use it on staging. If it is too confusing, we can iterate further.

PR #359, opened 2020-05-25
Style update by Zuri

navassouza:
Merge style update from Zuri PR: https://github.com/athenianco/athenian-webapp/pull/353


se7entyse7en:
I just noticed that had some conflicts in the spec file. When did it happen? I thought that they had been fixed.
Also, by looking at the commit history it seems that you used `merge` (twice) and solved the conflicts by committing, please use `rebase` next time. Anyway I don't wanna block this.

PR #360, opened 2020-05-25
Fix(ENG-935): Prevent the repo and dev filter to be open at the same time

navassouza:
Closes: https://athenianco.atlassian.net/browse/ENG-935


navassouza:
@dpordomingo, I'll address your comments in https://athenianco.atlassian.net/browse/ENG-951, 
I should have pushed them separately, my bad

PR #365, opened 2020-05-25
fix(ENG-889): Change the color of the GitHub handles from light grey to black when there is no full names

navassouza:
Closes: https://athenianco.atlassian.net/browse/ENG-889


se7entyse7en:
@navassouza needs conflict resolution

PR #366, opened 2020-05-26
feat(ENG-851): Apply the same hover effects than for the previous filters

navassouza:
Closes: https://athenianco.atlassian.net/browse/ENG-851


navassouza:
Screen Shot 2020-05-25 at 21 02 14

PR #374, opened 2020-05-26
Fix call to getTeams after interface change

se7entyse7en:


se7entyse7en:
At some point we could uniform the signature of the other functions as well to accept the api object instead of the token by taking it from the `useApi` custom hook.

PR #378, opened 2020-05-26
fix(ENG-970): Apply the same styling for the GitHub handles of the 2 listings

navassouza:
Closes: https://athenianco.atlassian.net/browse/ENG-970


se7entyse7en:
@navassouza No need for my approval for these kind of styling changes 👍

PR #386, opened 2020-05-28
Use only active days for CreatedPRs chart average KPI

dpordomingo:
fix [[ENG-761]] WIP: Pull Requests created through time

This will discard days when no PR was created when calculating the average

## notice
The Chart is still hidden at Prod; let me know if we should enable it, or feel free to PR it on your own.

[ENG-761]: https://athenianco.atlassian.net/browse/ENG-761


warenlg:
@dpordomingo feel free to enable the chart on prod.

PR #393, opened 2020-05-29
Fix copy in axis labels, legends and KPIs

warenlg:
Fix [ENG-992](https://athenianco.atlassian.net/browse/ENG-992)


dpordomingo:
Thank you for this initiative @warenlg

PR #400, opened 2020-06-01
Fix creation of instance id for QA env due to echo -n different implementations

se7entyse7en:

### DIFF
--- a/Makefile.testenv
+++ b/Makefile.testenv
@@ -3,7 +3,7 @@
 TEST_BRANCH ?=
 TEST_SERVICE_ACCOUNT ?=
 TEST_GCE_IMAGE ?= test-web-app
-TEST_GCE_INSTANCE ?= branch-$(shell sh -c "echo -n $(TEST_BRANCH) | md5sum | cut -c -32")
+TEST_GCE_INSTANCE ?= branch-$(shell sh -c "printf \"%s\" \"$(TEST_BRANCH)\" | md5sum | cut -c -32")
 TEST_GCE_ZONE ?= us-east1-c
 
 .PHONY: check-test-branch


se7entyse7en:
@lwsanty as discussed in Slack, tested and it worked. 👍

Makefile.testenv
```diff
@@ -3,7 +3,7 @@
 TEST_BRANCH ?=
 TEST_SERVICE_ACCOUNT ?=
 TEST_GCE_IMAGE ?= test-web-app
-TEST_GCE_INSTANCE ?= branch-$(shell sh -c "echo -n $(TEST_BRANCH) | md5sum | cut -c -32")
+TEST_GCE_INSTANCE ?= branch-$(shell sh -c "printf \"%s\" \"$(TEST_BRANCH)\" | md5sum | cut -c -32")
```
dpordomingo, line 6: I wonder if this one can fail in macOS too https://github.com/athenianco/athenian-webapp/blob/3eb21c12d3a6a109bd8d9d2b3688b3e834baaca7/Makefile.testenv#L48

se7entyse7en: Nope this is working fine as expected and printed in yellow 👍

se7entyse7en:
@lwsanty I'm merging this since we discussed it in Slack and tested and working fine 👍

se7entyse7en:
> Would it help to force the Makefile to use bash as done in athenianco/ci::Makefile.main

Sorry @dpordomingo I missed this. Anyway I wasn't aware that we're already doing it in main Makefile, then yes it makes sense.

dpordomingo:
I'm not sure it will work because I remember that macos is also doing another weird thing with the shell interpreter, I don't remember but it was related to using another interpreter because it symlinks one to another.
(I remember we had that issue in our former company, of symlinking `sh` to `zsh` or something like that in macos)

PR #404, opened 2020-06-02
Let God Users to impersonate other users from their Profile settings page

dpordomingo:
https://www.loom.com/share/7511504010a3465a81c9048a3d659371

## nice to have

Current API does not expose God capacities under `/user` endpoint, so it's checked directly from the Admin section directly trying to use it. If the user can not use it, the whole admin section is hidden from the Profile page.

It would be better if the God capacities are directly exposed by `/user` endpoint.

### DIFF
--- a/src/js/pages/settings/Profile.jsx
+++ b/src/js/pages/settings/Profile.jsx
@@ -1,51 +1,129 @@
-import React from 'react';
+import React, { useState, useEffect } from 'react';
+import classnames from 'classnames';
+import { useHistory} from 'react-router';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faGithub } from '@fortawesome/free-brands-svg-icons';
+import { faUserSecret } from '@fortawesome/free-solid-svg-icons';
 
 import defaultImage from 'images/default-user-image.png';
 
 import { useUserContext } from 'js/context/User';
 import { SettingsGroup, getOrg } from 'js/pages/Settings';
+import { useApi } from 'js/hooks';
+import { becomeUser, testGodUser } from 'js/services/api';
+import log from 'js/services/logger';
 
 export default () => {
-  const { user } = useUserContext();
+  const { user, godMode } = useUserContext();
   const org = getOrg(user);
 
   return (
-    <SettingsGroup title="Profile">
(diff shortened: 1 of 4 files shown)


warenlg:
This is very cool, very handy!
It would be amazing to have a breakdown listing the github ids together with the name and the github org.

dpordomingo:
You're right @warenlg, but we don't have such API endpoints yet.
If we would, it would be also an easy thing to be added.

src/js/pages/settings/Profile.jsx
```diff
@@ -1,51 +1,140 @@
...
+    (async () => {
+      setGodUserState(await testGodUser(api));
+    })();
+  }, [api, apiReady, isGodMode]);
+
+  const impersonate = async strategy => {
+    try {
+      if (!apiReady) {
+        throw new Error('Could not obtain an API client');
+      };
```
se7entyse7en, line 84: You can remove this check if you make `return null` also in case of `!apiReady`

src/js/pages/settings/Profile.jsx
```diff
@@ -1,51 +1,140 @@
...
+    } catch (e) {
+      log.fatal(`Could not impersonate user`, e);
+    }
+  };
+
+  const revert = async strategy => {
+    try {
+      if (!apiReady) {
+        throw new Error('Could not obtain an API client');
+      };
```
se7entyse7en, line 98: Same here

src/js/pages/settings/Profile.jsx
```diff
@@ -1,51 +1,140 @@
...
+
+      await becomeUser(api, null);
+      setTargetUserState('');
```
se7entyse7en, line 101: No need since you're refreshing the page soon after.

dpordomingo:
@se7entyse7en suggestions addressed by f52f6e4

rebased and ready to be merged once approved if there is no more suggestions.

se7entyse7en:
@dpordomingo What I meant is that you don't need to check for `apiReady` at all in those functions if you do:
```
if (!godUserState || !apiReady) return null;
```
anyway not a blocker 👍

dpordomingo:
You're right @se7entyse7en. Added and merged. Thanks!

PR #406, opened 2020-06-02
Public athenian

se7entyse7en:
Closes [DEV-162](https://athenianco.atlassian.net/browse/DEV-162).

This adds the possibility to see Athenian data in demo mode:
- to activate demo mode the user needs to visit `/demo`, it will set a cookie and will redirect to home showing the Athenian's data,
- during the demo, instead of the user avatar, the "Login" button is shown as if a user is not authenticated,
- to deactivate demo mode the user can visit `/undemo`,
- demo mode is deactivated automatically when going to `/login`,
- demo mode is deactivated also after 1h of "inactivity" (the cookie expires after 1h, but it refreshes on each page refresh or navigation, for example, navigating through the stages),
- in demo mode, a user cannot go to the settings page.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "node-sass": "^4.13.1",
     "popper.js": "^1.16.1",
     "react": "^16.12.0",
+    "react-cookie": "^4.0.3",
     "react-copy-to-clipboard": "^5.0.2",
     "react-dates": "^21.8.0",
     "react-dom": "^16.12.0",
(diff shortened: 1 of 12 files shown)


se7entyse7en:
@vmarkovtsev since the bearer token is not exposed anywhere on the webapp and the setting page cannot be accessed, it should be safe and there shouldn't be any need to prevent on api side to do write operations by @gkwillie.

eiso:
This looks perfect. Thank you for this. Once this is live we'll add it to our website.

PR #407, opened 2020-06-02
feat(DEV-188): Add ability to select range of PRs by Drag and select …

navassouza:
https://athenianco.atlassian.net/browse/DEV-188


eiso:
Is this ready to be tried on staging?

eiso:
Is this ready to be tried on staging?

navassouza:
@eiso, not yet. Had a chat with @vmarkovtsev yesterday and he confirmed that filtering the PRs using a date range in the frontend is not possible, a workaround he suggested was to re-fetch the list. Since it is not how we planned, I put this in hold so we can decide, a change in the api? Re-fetch indeed?

se7entyse7en:
@navassouza @vmarkovtsev @warenlg what we could do is to limit this for the overview page at list since we can use the `released` key that is a timestamp and filter on that.

WRT the api is it a big change to have also the corresponding timestamp along with each of the `properties` returned in the PR object?

warenlg:
> since we can use the released key that is a timestamp and filter on that.

I don't get it, we do have the timestamps fo the other properties/events
```
 "review_requested": "2020-06-02T12:54:40Z",
  "first_review": "2020-06-04T09:00:21Z",
  "approved": "2020-06-04T09:00:21Z",
  "review_comments": "0",
  "reviews": "1",
  "merged": "2020-06-04T09:00:42Z",
  "released": "2020-06-04T09:00:42Z",
```

vmarkovtsev:
> WRT the api is it a big change to have also the corresponding timestamp along with each of the properties returned in the PR object?

All of them are already reported as separate fields, as Waren noticed.

BTW @se7entyse7en we had a DM chat with Ricardo on Friday about this. We have to call `/filter/pull_requests` whenever the date range changes, and it is OK. What is not so OK is that the API will do quite a bit of work for an inner interval because the cache is not optimized for fast inner interval lookups. If the product thinks that we should have instant drag responses, I have to address that.

navassouza:
As per our discussion during daily:
- Have the highlight only on Overview for now
- Use 'released' date to filter PRs of Overview

PR #411, opened 2020-06-04
No data

dpordomingo:
required by [[DEV-189]] Display an informative visual when no data is available for the summary charts

-----

**note**: `insight` represents a single block of the webapp containing a chart and some optional data (KPIs and such), which is a candidate to be considered "Empty" under certain circumstances.

-----

Each `insight` is compounded of many pieces of information (KPIs, chart data, averages...), so there may be no a single way to define **WHEN** an `insight` must be considered "empty", nor **HOW** to show the "Empty" message; thus, each `insight` must provide its logic to be considered "empty".

For this initial iteration:
- an `insight` will be considered as "empty" **WHEN** there is no data to be plotted in the chart &mdash;no matter the rest of the info&mdash;.
- the way **HOW** an `insight` shows the "Empty" message is:
  - fading its content (`opacity: 0.2`), and
  - centering the same predefined "Empty" message in `insight` container.

In further iterations, each `insight` may define **WHEN** it must be considered "empty", and **HOW** to show the "Empty" message.

image

## TL;DR

`DataWidget` handles the "empty" criteria showing the Empty message for its plumbed data.
But in some circumstances, `DataWidget` is in charge of rendering many `insights` at once &mdash; like the thumbnails, or the stage insights&mdash;. In such circumstances, each `insight` block must be responsible of showing the "Empty" message following its own criteria.

## TODO (if really needed)

- define the "empty-default" state for each chart, to be shown under the "Empty" message. This PR does not handle that, so the chart is rendered directly by `react-vis` according to the data that it received.
- let each `insight` define **HOW** the "Empty" message is shown, &mdash;which can be also different per `insight`&mdash;.

[DEV-189]: https://athenianco.atlassian.net/browse/DEV-189


dpordomingo:
superseded by https://github.com/athenianco/athenian-webapp/pull/412

PR #412, opened 2020-06-04
"Empty" message when there is no data for an insight

dpordomingo:
required by [[DEV-189]] Display an informative visual when no data is available for the summary charts

-----

**note**: `insight` represents a single block of the webapp containing a chart and some optional data (KPIs and such), which is a candidate to be considered "Empty" under certain circumstances.

-----

Each `insight` is compounded of many pieces of information (KPIs, chart data, averages...), so there may be no a single way to define **WHEN** an `insight` must be considered "empty", nor **HOW** to show the "Empty" message; thus, each `insight` must provide its logic to be considered "empty".

For this initial iteration:
- an `insight` will be considered as "empty" **WHEN** there is no data to be plotted in the chart &mdash;no matter the rest of the info&mdash;.
- the way **HOW** an `insight` shows the "Empty" message is:
  - removing its content, but the card title, and
  - centering the same predefined "Empty" message in `insight` container.

In further iterations, each `insight` may define **WHEN** it must be considered "empty", and **HOW** to show the "Empty" message.

image

You can see below the `Overview Summary Metric` when there is no data:

image

## TL;DR

`DataWidget` handles the "empty" criteria showing the Empty message for its plumbed data.
But in some circumstances, `DataWidget` is in charge of rendering many `insights` at once &mdash; like the thumbnails, or the stage insights&mdash;. In such circumstances, each `insight` block must be responsible of showing the "Empty" message following its own criteria.

[DEV-189]: https://athenianco.atlassian.net/browse/DEV-189

### DIFF
--- a/src/js/components/DataWidget.jsx
+++ b/src/js/components/DataWidget.jsx
@@ -74,7 +74,7 @@ export default ({id, component, fetcher, plumber, globalDataIDs, config, propaga
                 console.error(error);
             } else {
                 widgetData = data;
-                status = widgetData ? READY : EMPTY;
+                status = widgetData && !widgetData.empty  ? READY : EMPTY;
             }
 
             console.log(`---> Rendering CHART: useEffect 1 | found ${error ? 'error' : 'data'}:`, error ? error : widgetData);
--- a/src/js/components/insights/Box.jsx
+++ b/src/js/components/insights/Box.jsx
@@ -2,6 +2,7 @@ import React from 'react';
 import classnames from 'classnames';
 
 import Info from 'js/components/ui/Info';
+import { NoData } from '../layout/Empty';
 /*
   Component for the boxes in the "Insights" tab of each stage of the pipeline.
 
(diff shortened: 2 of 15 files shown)


warenlg:
In case of no data, let's even remove the faded KPIs on the right of Zuri's visual. Otherzise, this sounds good to me. What about the summary charts?

dpordomingo:
I updated the PR desc replacing the screenshot with a new one showing both, `Stage Metric` and `Insights` when there is no data. I also added a new image including the `Overview Summary Metric` when there is no data in that case.

>_In case of no data, let's even remove the faded KPIs on the right of Zuri's visual.

Do you mean instead of fading the background, just drop it and show only the "Empty" message?

I mean, like the first `insight` instead of the second one.
image

Should we do the same for the stage and overview summary metrics?

warenlg:
I finally managed to test it with the pre-staging env, so I can give you proper review comments:
- Yes, let's remove the faded background like this:
83804729-dd90dc00-a6ae-11ea-8ff0-472879c0d6f1
- For some blocks, it seems the title is also faded. It should not:
Screenshot from 2020-06-05 13-24-57
- For the summary metrics, we should keep the white background-color, remove the faded KPIs on the left and keep the **Review** title in yellow.
Screenshot from 2020-06-05 13-28-36
- In the future, we should also force the blocks with no data to have a fixed height.

PR #441, opened 2020-06-11
fix(DEV-248): The developers are sent twice when filtering

navassouza:
Closes: https://athenianco.atlassian.net/browse/DEV-248

Send unique dev entries from filter, duplication was causing errors in graphs


src/js/services/api/index.js
```diff
@@ -234,7 +234,7 @@ export const fetchFilteredPRs = async (
     exclude_inactive
 ) => {
     filter.repositories = filter.repositories || [];
-    filter.developers = _(filter.developers || []).map(v => v.login).value();
+    filter.developers = _(filter.developers || []).uniqBy('login').map(v => v.login).value();
```
dpordomingo, line 237: I wonder if we could move that logic into the filter, so it would be needed only once; I mean, couldn't we ensure that `filter.developers` will contain only unique developers?

dpordomingo: Otherwise, we could face this same problem in some other places.

navassouza: We need the duplication to have the filters work when a dev is in more than 1 team. Moved the uniqBy out of the api methods and into the where we access the context and get the list of devs

dpordomingo: I see the necessity of having duplicated users in the internal state of the `UsersMultiSelect` component, but why expose it to the outside?
Imo the internal state of that component should not be exposed, but its deduplicated version, so wherever it would be used, we would already have a clean copy of the filtered (and deduplicated) users.

dpordomingo: Anyhow, if I must choose between deduplicating in the API, or in the methods accessing the contributors context, I'd do it in the API methods because it would be more reusable.

dpordomingo: For the record. I guess my concern is related to this `TODO` from @se7entyse7en 
https://github.com/athenianco/athenian-webapp/blob/master/src/js/pages/pipeline/Filters/index.jsx#L82

navassouza:
Merging this PR, the concerned raised by @dpordomingo will be address in another PR. To be discussed with @se7entyse7en as there's a TODO comment related to this issue.

PR #455, opened 2020-06-16
Feat(ENG-959): Apply the same style of the 'Repositories' and 'Users' filters to the 'Filter by status'

navassouza:
Cloes: https://athenianco.atlassian.net/browse/ENG-959
Apply same style to Pr Table filter


src/js/components/pipeline/PullRequests.jsx
```diff
@@ -112,6 +113,47 @@ export default ({ stage, data, status }) => {
...
+                  )
+                }
+              }}
+              styles={{
+                control: (provided, state) => ({
+                  ...provided,
+                  borderRadius: 0,
+                  boxShadow: 'none',
+                  borderColor: state.isFocused
+                  ? '#ffd188'
```
dpordomingo, line 131: why not using plain css and already existent color variables, styles, and such?

navassouza: react-select has this interface for inner components which helps to customize it, as you'll see each css object is assigned to an inner component it's not css in the loose. In order to override with plain css it we'd need to 1) increase specificity, 2) use those auto-generated classes from react-select (e.g.: controls-zDtafdf), I'd rather use their approach using `emotion`. We have used it in the filters for Repos and Developers. One thing I agree is move this code to be shareable, just like the other filters

PR #460, opened 2020-06-16
Consider "all" reposet as the default

vmarkovtsev:


src/js/services/api/index.js
```diff
@@ -89,13 +89,15 @@ export const getUserWithAccountRepos = async token => {
       return { id: Number(accountID), isAdmin, reposets: [] };
     }
 
-    const reposetsContent = await Promise.all(reposets.map(async reposetListItem => {
+    const reposetsContent = await reposets.reduce(async (map, reposetListItem) => {
```
dpordomingo, line 92: this is creating a promise of a dictionary of promises of reposets (that's because we needed the `Promise.all`)

dpordomingo, line 92: [`Array.reduce`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce) `reducer` use to use `accumulator` or `acc` as the first parameter name, which is more clear about how the `reducer` internally works.

src/js/services/api/index.js
```diff
@@ -89,13 +89,15 @@ export const getUserWithAccountRepos = async token => {
...
+      map[reposetWithName.name] = {
         id: reposetListItem.id,
-        repos: reposetWithName.items,
-      }
-    }));
+        name: reposetWithName.name,
+        repos: reposetWithName.items
+      };
+      return map;
+    });
```
dpordomingo, line 100: The `initialValue` is highly recommended to be defined in [`Array.reduce`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce) in order to avoid:
- using the first array item as the initial value for the `accumulator` in the first pass,
- avoiding `reduce` function to throw an error when working in an empty array (because there is no first value to initialize the `accumulator` when there is no `initialValue`)


(from docs)
image

src/js/services/api/index.js
vmarkovtsev: I was guided by [StackOverflow](https://stackoverflow.com/a/41243567/69708) here.
Why on earth in JS does `await` return a `Promise`?

PR #461, opened 2020-06-17
All reposet default

dpordomingo:
supersedes [#460] Consider "all" reposet as the default


vmarkovtsev:
Why not having it as a map?

dpordomingo:
We should respect the current shape of the returned `user.defaultAccount`

```js
{
  id: 1,
  isAdmin: true,
  reposets: [
    {
      id: 1,
      name: "all",
      repos: (2) ["github.com/src-d/gitbase", "github.com/src-d/go-git"]
      items_count: 2,
      …
    },
  ]
}
```

which expects an array of named reposets under `reposets` key  instead of a dictionary of reposets.

We could for sure review and refactor current code to expect a dictionary, but I don't see the benefit especially now, that we're not using named reposets at all.

vmarkovtsev:
I found the only place where we had to change from array to dict, which was where we filter by `"all"`. Is there anything else?

